PR #21, opened 2022-10-20
resolve i-test in phase:pre-integration-test, generic itest.bndrun

stbischof:


timothyjward (commented):
The bndrun changes look good, but do we want a `writeOnChanges` as part of every build? I think the resolve step should be validation only unless requested by the user with a property. 

Thoughts?

stbischof:
I did it like https://github.com/osgi/osgi-test/blob/f97a098e78406199927894ef44f68e9ab1a92f5a/pom.xml#L178

It helps me to find faster if something does not really resolves but still would run because of the old resolve result.

timothyjward:
> It helps me to find faster if something does not really resolves but still would run because of the old resolve result.

I agree, but validation would also do that, as well as flagging unexpected changes in dependency. 

Just defaulting  writeOnChanges to false and failOnChanges to true, with a property to set writeOnChanges to true if a manual resolve is needed.

stbischof:
removed the phase change, have to play a bit more to see hot it works

timothyjward (approved):

PR #24, opened 2022-10-20
Simplify with @InjectMocks, @ParameterizedTest, descriptive fields

stbischof:
remove unused `@Mock Resource resource;`

### DIFF
--- a/prototype/core/impl/src/test/java/org/eclipse/sensinact/prototype/command/impl/GatewayThreadImplTest.java
+++ b/prototype/core/impl/src/test/java/org/eclipse/sensinact/prototype/command/impl/GatewayThreadImplTest.java
@@ -20,7 +20,6 @@ import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 
 import org.eclipse.emf.common.util.URI;
-import org.eclipse.emf.ecore.resource.Resource;
 import org.eclipse.emf.ecore.resource.ResourceSet;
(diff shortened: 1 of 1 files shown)


prototype/core/impl/src/test/java/org/eclipse/sensinact/prototype/command/impl/GatewayThreadImplTest.java
```diff
@@ -69,28 +65,31 @@ void teardown() {
         thread.deactivate();
     }
 
-    @Test
-    void testExecute() throws Exception {
+    private static final int DELAY = 100;
+    private static final int TEST_VALUE = 5;
+
+    @ParameterizedTest
+    @ValueSource(ints = { 1, 10, 100, 1000 })
```
timothyjward, line 72: I don't think that 1 and 10 are good values to use here. If the CI server is heavily loaded then we are likely to see timing related failures and "flaky tests" for the line:
 
    assertFalse(result.isDone())

I'd prefer not to remove this test expression as it guarantees that the command is run asynchronously, hence removing the very short time delays.

timothyjward (changes requested):

prototype/core/impl/src/test/java/org/eclipse/sensinact/prototype/command/impl/GatewayThreadImplTest.java
stbischof: ok

PR #28, opened 2022-10-26
Further PR checking improvements

timothyjward:
CI build and stale issue management


stbischof:
what about matrix build?
https://github.com/adoptium/temurin-build/blob/ce8eca7591182a7d05adf1a25d7feac1059c1ef2/.github/workflows/build.yml#L20

timothyjward:
> what about matrix build?
> [example](https://github.com/adoptium/temurin-build/blob/ce8eca7591182a7d05adf1a25d7feac1059c1ef2/.github/workflows/build.yml#L20)

This would be good as another contribution to the workflow

PR #37, opened 2022-11-01
fix pre-editconfig formats

stbischof:
Signed-off-by: Stefan Bischof <stbischof@bipolis.org>


stbischof:
@timothyjward here is an example how editorconfig-check fails on javadocs

```
/**
 * is
 *\
```

```
/**
* expected
*\
```

PR #40, opened 2022-11-02
Use super-linter for editorconfig validation

timothyjward:


stbischof:
super-linter is a good idea!!!

PR #42, opened 2022-11-03
Added API & DTOs for REST northbound endpoint

tcalmant:


tcalmant:
The ip-check error comes from the `org.osgi.service.jakartars 2.0.0-SNAPSHOT` dependency being unknown to ClearlyDefined.
This artifact comes from the https://oss.sonatype.org/ repository.

PR #53, opened 2022-11-18
implemented persistence of models between restarts

juergen-albert:
Signed-off-by: Juergen Albert <j.albert@data-in-motion.biz>


prototype/.editorconfig
```diff
@@ -12,6 +12,8 @@ indent_size = 4
 indent_style = space
 trim_trailing_whitespace = true
 
+[core/models/src/*]
```
timothyjward, line 15: I think this was added by mistake. It doesn't change any settings.

prototype/core/impl/pom.xml
```diff
@@ -53,6 +53,10 @@
       <groupId>org.geckoprojects.emf</groupId>
       <artifactId>org.gecko.emf.osgi.component</artifactId>
     </dependency>
+    <dependency>
+      <groupId>biz.aQute.bnd</groupId>
+      <artifactId>aQute.libg</artifactId>
+    </dependency>
```
timothyjward, line 59: I don't think we should be adding dependencies on bnd libraries unless we really need them. This looks like it's only there for `Exceptions.duck.

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/NexusImpl.java
```diff
@@ -64,9 +76,83 @@ public NexusImpl(ResourceSet resourceSet, SensiNactPackage sensinactPackage,
...
+        if (new File(fileName).exists()) {
+            try {
+                resource.load(null);
+                if (!resource.getContents().isEmpty()) {
+                    EPackage defaultPackage = (EPackage) resource.getContents().get(0);
+                    resource.setURI(URI.createURI(defaultPackage.getNsURI()));
+                    return Optional.of(defaultPackage);
+                }
+            } catch (IOException e) {
+                Exceptions.duck(e);
```
timothyjward, line 99: Ducking exceptions is (in my opinion) bad practice and we shouldn't do it. It will also allow us to remove the bnd libg dependency.

Another positive benefit will be that the constructor has to make a call about whether a failure to load the default package should blow up the creation of the NexusImpl or not (it does at the moment) and it will force some logging.

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/NexusImpl.java
```diff
@@ -64,9 +76,83 @@ public NexusImpl(ResourceSet resourceSet, SensiNactPackage sensinactPackage,
...
 
-        defaultPackage = EMFUtil.createPackage("base", DEFAULT_URI, "sensinactBase", this.resourceSet);
+        defaultPackage = packageOptional
+                .orElseGet(() -> EMFUtil.createPackage("base", DEFAULT_URI, "sensinactBase", this.resourceSet));
         packageCache.put(DEFAULT_URI_OBJECT, defaultPackage);
+        loadInstances();
+    }
+
+    private Optional<EPackage> loadDefaultPackage(String fileName) {
+        Resource resource = resourceSet.createResource(URI.createFileURI(fileName));
```
timothyjward, line 89: This should take a `Path`, which is what we have when we call it

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/NexusImpl.java
```diff
@@ -64,9 +76,83 @@ public NexusImpl(ResourceSet resourceSet, SensiNactPackage sensinactPackage,
...
+        if (new File(fileName).exists()) {
```
timothyjward, line 90: `Files.isRegularFile(Path)` would be better

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/NexusImpl.java
```diff
@@ -64,9 +76,83 @@ public NexusImpl(ResourceSet resourceSet, SensiNactPackage sensinactPackage,
...
+                Exceptions.duck(e);
+            }
+        }
+        return Optional.empty();
+    }
+
+    private void loadInstances() {
+
+        Path instancesPath = Paths.get(INSTANCES);
+        if (instancesPath.toFile().exists()) {
```
timothyjward, line 108: `Files.isDirectory(Path)` would be better

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/NexusImpl.java
```diff
@@ -64,9 +76,83 @@ public NexusImpl(ResourceSet resourceSet, SensiNactPackage sensinactPackage,
...
+            try {
+                Files.walk(instancesPath).forEach(this::load);
+            } catch (IOException e) {
+                Exceptions.duck(e);
+            }
```
timothyjward, line 113: Don't duck the exception. Handle it and make a decision to fail the whole load, or log something

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/NexusImpl.java
```diff
@@ -64,9 +76,83 @@ public NexusImpl(ResourceSet resourceSet, SensiNactPackage sensinactPackage,
...
+        }
+    }
+
+    private void load(Path path) {
+        File file = path.toFile();
```
timothyjward, line 118: Not needed

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/NexusImpl.java
```diff
@@ -64,9 +76,83 @@ public NexusImpl(ResourceSet resourceSet, SensiNactPackage sensinactPackage,
...
+        try {
+            if (file.isDirectory()) {
```
timothyjward, line 120: `Files.isDirectory(Path)`

prototype/pom.xml
```diff
@@ -287,6 +286,11 @@
         <artifactId>jackson-databind</artifactId>
         <version>${jackson.version}</version>
       </dependency>
+      <dependency>
+        <groupId>biz.aQute.bnd</groupId>
+        <artifactId>aQute.libg</artifactId>
+        <version>${bnd.version}</version>
+      </dependency>
```
timothyjward, line 293: No libg compile dependency needed

prototype/core/models/src/main/resources/model/sensinact.genmodel
```diff
@@ -8,7 +8,8 @@
   <foreignModel>sensinact.ecore</foreignModel>
   <genPackages prefix="SensiNact" basePackage="org.eclipse.sensinact.model" resource="XMI"
       disposableProviderFactory="true" literalsInterface="false" ecorePackage="sensinact.ecore#/">
-    <genDataTypes ecoreDataType="sensinact.ecore#//EInstant"/>
```
timothyjward, line 11: This is a different change and should be in a different PR

timothyjward (changes requested):

prototype/.editorconfig
juergen-albert: This excludes any checks on the emf stuff. The automatically written ecore and genmodel and the generated code does not conform to the settings of editorconfig. If I don't do it like this, it fails all the time, as the code tends to be regenerated on every build.

timothyjward:
Hi Jürgen, the overall thrust of this PR is great, but it's quite large and appears to be doing two separate things

* Changing the data type for a timestamp in the model to be an instant
* Adding a persistent store for models

I think the PR needs to be split into those two things, and we need to work on why there are so many changes in the generated code (perhaps the generated code shouldn't be checked in?).

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/NexusImpl.java
juergen-albert: Makes sense. The Main question would be, what we should do. If the load of the old Data fails for whatever reason, should we just log it as an error and keep on going?

timothyjward: > should we just log it as an error and keep on going?

Initially yes. I think that it may eventually depend on the error type (e.g. a versioning error might indicate a failed migration in upgrade), but if the file is garbage then we probably want a big warning in the log and to continue without it.

juergen-albert:
Most of the generated Code changes are due to a first run of the Code Formatter on it.

timothyjward:
> Most of the generated Code changes are due to a first run of the Code Formatter on it.

I noticed that it has added tabs, which is inconsistent with the rest of the Java sources...

timothyjward (approved):

PR #54, opened 2022-11-21
untracked and ignored generated code

juergen-albert:
Signed-off-by: Juergen Albert <j.albert@data-in-motion.biz>


timothyjward (approved):

timothyjward:
Removing the generated code from the git repository will definitely simplify development going forward.

PR #59, opened 2022-11-23
Gateway constructor injection

juergen-albert:


prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/command/impl/GatewayThreadImpl.java
```diff
@@ -39,15 +39,8 @@
 @Component
 public class GatewayThreadImpl extends Thread implements GatewayThread {
 
-    @Reference
     TypedEventBus typedEventBus;
```
timothyjward, line 42: `private final TypedEventBus typedEventBus;`

timothyjward: I assume that this was the main reason for constructor injection?

PR #63, opened 2022-11-24
Fixes on SensorThings

tcalmant:
* Use internal GeoJSON library
* Fixed filters behaviour when related argument is missing
* Don't use `@Context` to find the registered `ObjectMapper` and `SensinactSession` providers
* Normalized conversion of `admin/location` to a `GeoJsonObject`

### DIFF
--- a/prototype/northbound/sensorthings/rest.api/src/main/java/org/eclipse/sensinact/sensorthings/sensing/rest/filters/CountFilter.java
+++ b/prototype/northbound/sensorthings/rest.api/src/main/java/org/eclipse/sensinact/sensorthings/sensing/rest/filters/CountFilter.java
@@ -40,7 +40,7 @@ public class CountFilter implements ContainerRequestFilter, ContainerResponseFil
             throws IOException {
         Boolean addCount = (Boolean) requestContext.getProperty(COUNT_PROP);
 
-        if(addCount) {
+        if(addCount != null && addCount) {
             Object entity = responseContext.getEntity();
             if(entity instanceof ResultList) {
                 ResultList<?> resultList = (ResultList<?>) entity;
@@ -53,7 +53,7 @@ public class CountFilter implements ContainerRequestFilter, ContainerResponseFil
     public void filter(ContainerRequestContext requestContext) throws IOException {
         boolean addCount = false;
 
-        List<String> list = requestContext.getUriInfo().getQueryParameters().get("$count");
+        List<String> list = requestContext.getUriInfo().getQueryParameters().getOrDefault("$count", List.of());
         if(list.size() > 1) {
             requestContext.abortWith(Response
                     .status(Status.BAD_REQUEST)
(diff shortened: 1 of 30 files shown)


prototype/northbound/sensorthings/rest.api/src/main/java/org/eclipse/sensinact/sensorthings/sensing/rest/filters/OrderByFilter.java
```diff
@@ -43,17 +45,26 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
...
+            List<Self> sortedList = new ArrayList<>(resultList.value);
+            sortedList.sort(comparator);
+            resultList.value = List.copyOf(sortedList);
         }
     }
 
     @Override
     public void filter(ContainerRequestContext requestContext) throws IOException {
         List<String> list = requestContext.getUriInfo().getQueryParameters().get("$orderby");
+        if (list == null) {
```
timothyjward, line 65: This behaviour is wrong. In the absence of a sort property we should default to sorting by id as per the spec

prototype/northbound/sensorthings/rest.api/src/main/java/org/eclipse/sensinact/sensorthings/sensing/rest/filters/OrderByFilter.java
```diff
@@ -43,17 +45,26 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
             throws IOException {
         @SuppressWarnings("unchecked")
         Comparator<Object> comparator = (Comparator<Object>) requestContext.getProperty(ORDERBY_PROP);
+        if (comparator == null) {
```
timothyjward, line 48: There should always be a comparator set to sort the results.

prototype/northbound/sensorthings/rest.api/src/main/java/org/eclipse/sensinact/sensorthings/sensing/rest/filters/CountFilter.java
```diff
@@ -40,7 +40,7 @@ public void filter(ContainerRequestContext requestContext, ContainerResponseCont
             throws IOException {
         Boolean addCount = (Boolean) requestContext.getProperty(COUNT_PROP);
 
-        if(addCount) {
+        if(addCount != null && addCount) {
```
timothyjward, line 43: Add count should not be null, see the bug below

prototype/northbound/sensorthings/rest.api/src/main/java/org/eclipse/sensinact/sensorthings/sensing/rest/filters/PropFilterImpl.java
```diff
@@ -38,6 +38,12 @@ public void aroundWriteTo(WriterInterceptorContext context) throws IOException,
         Object entity = context.getEntity();
 
         String propName = uriInfo.getPathParameters().getFirst("prop");
+        if(propName == null) {
+            // No property filter
```
timothyjward, line 42: Isn’t this an error? The prop filter name binding is supposed to be used with a path parameter that cannot be null, so I’m not sure how this code is reached. If it is reached we should abort with a failure.

tcalmant: Agreed. This case only happens if the annotated method doesn't have a "prop" parameter.

PR #83, opened 2022-12-09
Implementation of the device factory handler

tcalmant:
Implements the device factory handler.
Requires #81 to be merged to work.

### DIFF
--- a/prototype/southbound/device-factory/device-factory-core/src/main/java/org/eclipse/sensinact/gateway/southbound/device/factory/impl/FactoryParserHandler.java
+++ b/prototype/southbound/device-factory/device-factory-core/src/main/java/org/eclipse/sensinact/gateway/southbound/device/factory/impl/FactoryParserHandler.java
@@ -1,0 +1,727 @@
+/*********************************************************************
+* Copyright (c) 2022 Contributors to the Eclipse Foundation.
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which is available at https://www.eclipse.org/legal/epl-2.0/
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*   Kentyou - initial implementation
+**********************************************************************/
+package org.eclipse.sensinact.gateway.southbound.device.factory.impl;
+
+import java.time.Instant;
+import java.time.LocalDate;
+import java.time.OffsetDateTime;
+import java.time.OffsetTime;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoField;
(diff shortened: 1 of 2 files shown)


prototype/southbound/device-factory/device-factory-core/src/main/java/org/eclipse/sensinact/gateway/southbound/device/factory/impl/FactoryParserHandler.java
```diff
@@ -0,0 +1,709 @@
...
+    }
+
+    /**
+     * Looks for a location in the given record
+     *
+     * @param record       Record to read
+     * @param placeholders Defined mapping placeholders
+     * @return The parsed location as a GeoJSON string or null
+     */
+    private String computeLocation(final IDeviceMappingRecord record, final Map<String, RecordPath> placeholders) {
```
timothyjward, line 410: I think that this should be returning a GeoJson Object, not a String. That's what the value of the admin/location should be.

prototype/southbound/device-factory/device-factory-core/src/main/java/org/eclipse/sensinact/gateway/southbound/device/factory/impl/FactoryParserHandler.java
```diff
@@ -0,0 +1,709 @@
...
+
+        if (svcRefs == null || svcRefs.isEmpty()) {
+            throw new MissingParserException(String.format("Parser ID '%s' is missing", parserId));
+        }
+
+        return svcRefs.stream().sorted(Comparator.naturalOrder()).findFirst().get();
+    }
+
+    @Override
+    public void handle(final SensiNactSession session, final DeviceMappingConfigurationDTO configuration,
```
timothyjward, line 144: I think that the `SensiNactSession` can be removed from here now? If not then what user does it act as?

timothyjward (approved):

prototype/southbound/device-factory/device-factory-core/src/main/java/org/eclipse/sensinact/gateway/southbound/device/factory/impl/FactoryParserHandler.java
tcalmant: Agreed, will do

prototype/southbound/device-factory/device-factory-core/src/main/java/org/eclipse/sensinact/gateway/southbound/device/factory/impl/FactoryParserHandler.java
tcalmant: Agreed

PR #90, opened 2023-01-04
MQTT device factory

tcalmant:
Links the MQTT listener to the device factory


prototype/southbound/mqtt/mqtt-device-factory/src/main/java/org/eclipse/sensinact/gateway/southbound/mqtt/factory/MqttDeviceFactoryHandler.java
```diff
@@ -0,0 +1,146 @@
...
+
+    @Override
+    public void onMqttMessage(String handlerId, String topic, IMqttMessage message) {
+        if (allowedHandlerId != null && !allowedHandlerId.equals(handlerId)) {
+            // Ignore message
+            return;
+        }
+
+        // TODO: define the correct session
+        final SensiNactSession session = sessionManager.getDefaultSession("user");
```
timothyjward, line 139: This doesn't feel quite right. Should we be using push updates and no session? Probably one to discuss in Grenoble as part of the security discussion.

prototype/southbound/mqtt/mqtt-device-factory/src/test/resources/configs/config.json
```diff
@@ -0,0 +1,45 @@
+{
```
timothyjward, line 1: Have you considered using https://github.com/osgi/osgi-test/tree/main/org.osgi.test.junit5.cm instead of the configurator? It keeps the configuration closed to the test code that's using it.

prototype/southbound/mqtt/mqtt-device-factory/src/main/java/org/eclipse/sensinact/gateway/southbound/mqtt/factory/MqttDeviceFactoryHandler.java
tcalmant: This is a relic from the previous Device Factory Handler API, which now uses the `PrototypePush` API. I'm updating the call here.

prototype/southbound/mqtt/mqtt-device-factory/src/test/resources/configs/config.json
tcalmant: Thanks for the example, I've updated the PRs

timothyjward (approved):

PR #95, opened 2023-01-10
HTTP device factory implementation

tcalmant:
Based on Jetty 11 clients using shared resources.
Supports custom headers, Basic Auth, SSL client auth and SSL server certificate verification with custom trust store.
Sends the HTTP response to the device factory handler.


timothyjward (approved):

tcalmant:
Also fixed an issue with initial timestamps of resources from the admin service

PR #105, opened 2023-01-23
Implementation of feature configurations

tcalmant:


prototype/distribution/launcher/src/main/java/org/eclipse/sensinact/gateway/launcher/ConfigurationManager.java
```diff
@@ -196,4 +223,69 @@ private void createConfig(String pid, Hashtable<String, Object> value) throws IO
...
+        if(configProperties.isEmpty()) {
+            // Generate the mandatory properties
+            configResource.getProperties().put(":configurator:resource-version", 1);
+            // TODO: use the bundle context to get those?
+            configResource.getProperties().put(":configurator:symbolic-name", "org.eclipse.sensinact.gateway.launcher");
+            configResource.getProperties().put(":configurator:version", "0.0.1");
+        }
+
+        // Remove configuration
+        if (removedConfigurations != null) {
```
timothyjward, line 252: We should discuss whether this is the right thing to do. Possibly we should only remove them if they haven't been changed by the user? I might get a bit annoyed if I hand edited a config file and the section got automatically deleted.

prototype/distribution/launcher/src/main/java/org/eclipse/sensinact/gateway/launcher/ConfigurationManager.java
```diff
@@ -196,4 +223,69 @@ private void createConfig(String pid, Hashtable<String, Object> value) throws IO
...
+            removedConfigurations.stream().forEach(allConfigurations::remove);
+        }
+
+        // Merge new configurations with the current values
+        if (newConfigurations != null) {
+            final Map<String, Hashtable<String, Object>> mergedConfigs = new Hashtable<>(newConfigurations.size());
```
timothyjward, line 258: We should be careful about ordering here. In general we use LinkedHashMap instances to preserve key-insertion order and avoid re-ordering the entries in the config file unnecessarily.

prototype/distribution/launcher/src/main/java/org/eclipse/sensinact/gateway/launcher/ConfigurationManager.java
```diff
@@ -63,7 +66,10 @@ public class ConfigurationManager {
 
     ExecutorService executor;
 
-    private Path configFile;
+    Path configFile;
+
+    private boolean expectedInterruption = false;
```
timothyjward, line 71: This needs to be volatile (or better yet an AtomicBoolean) as the signal comes from a different thread and (as far as I am aware) Thread.interrupt() doesn't trigger a memory barrier and force a memory update in either thread.

timothyjward (changes requested):

prototype/distribution/launcher/src/main/java/org/eclipse/sensinact/gateway/launcher/ConfigurationManager.java
```diff
@@ -108,12 +114,25 @@ void stop() {
...
             try {
                 key = watchService.take();
-            } catch (ClosedWatchServiceException | InterruptedException e) {
+            } catch (ClosedWatchServiceException e) {
                 LOGGER.error("No longer able to monitor the configuration file {}", configFile, e);
                 return;
+            } catch (InterruptedException e) {
+                if (expectedInterruption) {
+                    LOGGER.debug("Forced configuration reload");
+                    Thread.interrupted();
```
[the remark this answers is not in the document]
tcalmant: OK

prototype/distribution/launcher/src/main/java/org/eclipse/sensinact/gateway/launcher/ConfigurationManager.java
tcalmant: OK. Using an AtomicBoolean is better as we can do a getAndSet while handling the interruption

prototype/distribution/launcher/src/main/java/org/eclipse/sensinact/gateway/launcher/ConfigurationManager.java
tcalmant: OK.

prototype/distribution/launcher/src/main/java/org/eclipse/sensinact/gateway/launcher/ConfigurationManager.java
tcalmant: This needs some discussion indeed, as not removing a previous configuration means keeping potentially removed properties

prototype/distribution/launcher/src/main/java/org/eclipse/sensinact/gateway/launcher/ConfigurationManager.java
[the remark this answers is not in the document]
timothyjward: Having checked further, the code throwing the `InterruptedException` is supposed to clear the interrupt so we shouldn't here. Clearing it again risks missing a natural "stop" that happens just after a reload call.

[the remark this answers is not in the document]
tcalmant: OK, understood. The `InterruptedException` thrown by `WatchService.take()` is in fact a propagation of the one thrown by `Condition.await()`.
As indicated in its [javadoc](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/locks/Condition.html#await()), the interruption flag is cleared in that case.

timothyjward (approved):

PR #106, opened 2023-01-25
Require that admin/location is a GeoJsonObject

timothyjward:
The Admin service location resource should always be valid Geo Json. We can require this by using a GeoJsonObject as the type of the resource.

Signed-off-by: Tim Ward <timothyjward@apache.org>

### DIFF
--- a/prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/ModelNexus.java
+++ b/prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/ModelNexus.java
@@ -34,6 +34,8 @@ import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.EAnnotation;
 import org.eclipse.emf.ecore.EAttribute;
 import org.eclipse.emf.ecore.EClass;
+import org.eclipse.emf.ecore.EClassifier;
+import org.eclipse.emf.ecore.EDataType;
 import org.eclipse.emf.ecore.EPackage;
 import org.eclipse.emf.ecore.EStructuralFeature;
 import org.eclipse.emf.ecore.resource.Resource;
@@ -309,10 +311,14 @@ public class ModelNexus {
         // Allow an update if the resource didn't exist or if the update timestamp is
         // equal to or after the one of the current value
         if (metadata == null || !metadata.getTimestamp().isAfter(timestamp)) {
-            service.eSet(resourceFeature, data);
+            EClassifier resourceType = resourceFeature.getEType();
+            if (data == null || resourceType.isInstance(data)) {
+                service.eSet(resourceFeature, data);
+            } else {
+                service.eSet(resourceFeature, EMFUtil.convertToTargetType(resourceType, data));
+            }
             accumulator.resourceValueUpdate(modelName, providerName, serviceFeature.getName(),
-                    resourceFeature.getName(), resourceFeature.getEType().getInstanceClass(), oldValue, data,
-                    timestamp);
+                    resourceFeature.getName(), resourceType.getInstanceClass(), oldValue, data, timestamp);
         } else {
             return;
         }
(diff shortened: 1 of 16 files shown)


timothyjward:
@juergen-albert I don't know if there is a more efficient way to do the to/from String transform

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/ModelNexus.java
```diff
@@ -309,10 +311,14 @@ private void setData(Provider provider, ModelTransaction transaction, String mod
         // Allow an update if the resource didn't exist or if the update timestamp is
         // equal to or after the one of the current value
         if (metadata == null || !metadata.getTimestamp().isAfter(timestamp)) {
-            service.eSet(resourceFeature, data);
+            EClassifier resourceType = resourceFeature.getEType();
+            if (data == null || resourceType.isInstance(data)) {
+                service.eSet(resourceFeature, data);
+            } else {
+                service.eSet(resourceFeature, EMFUtil.convertToTargetType(resourceType, data));
```
juergen-albert, line 318: I'm not sure about this part. The EMFUtil methode introdcued here tries to check null again and set the default value in such a case. This will never apply as the null has already been checked.

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/emf/EMFUtil.java
```diff
@@ -175,4 +178,17 @@ public static int getContainerVersion(EStructuralFeature feature) {
...
+    public static Object convertToTargetType(EClassifier targetType, Object o) {
+
+        if (o == null) {
+            return targetType.getInstanceClass().isPrimitive() ? targetType.getDefaultValue() : o;
+        } else {
+            EClassifier type = typeMap.get(o.getClass());
+            if (type == null) {
+                throw new IllegalArgumentException("Unknown data type " + o.getClass());
+            }
+            String string = type.getEPackage().getEFactoryInstance().convertToString((EDataType) type, o);
```
juergen-albert, line 190: This seems like an unneceesary step or do we want to use this to create a deep copy of the underlying object?

timothyjward: Are you suggesting that we should do a type check and short circuit at the start?

timothyjward: The main reason for all of this was to support setting admin/location as a String and parsing it appropriately, but also to allow for simple Long/Integer processing

juergen-albert: This seems like an unneceesary step or do we want to use this to create a deep copy of the underlying object?

juergen-albert: Sorry, my mistake. I missed the distinction between type and targetType. In this light it makes sense.

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/emf/EMFUtil.java
```diff
@@ -17,6 +17,7 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
```
juergen-albert, line 20: Seems to be unused.

juergen-albert:
```suggestion
```

juergen-albert:
LGTM

Issue #109, opened 2023-01-26
Resource values are lost when adding a new resource


tcalmant:
When creating new resources calling `ModelNexus.handleDataUpdate()`, all the resources the service was already containing have their values removed (set to null).

I have added a [test](https://github.com/tcalmant/org.eclipse.sensinact.gateway/blob/thomas/value-bug/prototype/core/impl/src/test/java/org/eclipse/sensinact/prototype/model/nexus/impl/NexusTest.java#L242-L265) to show the error.

I suspect the issue comes from the [attribute version check](https://github.com/tcalmant/org.eclipse.sensinact.gateway/blob/thomas/value-bug/prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/emf/SensinactCopier.java#L47) in `SensinactCopier`

tcalmant:
@juergen-albert, I might need some help as I don't understand the behaviour of EMF copy around there.
There seems to be an incorrect copy of the `eStorage` field of the `ProviderImpl` when calling the `SensinactCopier` when the service state is new in [ModelNexus.updateInstances()](https://github.com/tcalmant/org.eclipse.sensinact.gateway/blob/thomas/value-bug/prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/ModelNexus.java#L443)

juergen-albert:
*facepalm* I see the issue. I'm on it.

PR #111, opened 2023-01-26
Emf 2.33  and fix for #109

juergen-albert:


juergen-albert:
@timothyjward and @tcalmant The fix to #109 is the update to the latest Version of EMF, where the new mechanism of Ed takes care of this issue. Fixing the Copier would have been a bit of a hassle in comparison. 

EMF is part of the Eclipse Release train and is thus not released and on maven central yet. Not sure how IP Dash will deal with it.. I have to update the geckoEMF Codegenerator with the new Version, which I can't publish to maven central until the release is out. Till then we would have to live with the following
* generated Code checked in
* code generated is commented out and need to be put back in place when model changes need to take effekt
* If code is regenerated, we need to manually run "Organize Imports" on as the current version of the generator misses an import

timothyjward:
Are we ok to merge this? I know that the checked in code isn't ideal, but we could do to keep working with this update. Hopefully the quarterly release train will mean that we can switch back in a month or two.

PR #119, opened 2023-02-07
Implement programmatic model building

timothyjward:
Use the programmatic API to create models and instances at runtime

Signed-off-by: Tim Ward <timothyjward@apache.org>


prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/ModelNexus.java
```diff
@@ -469,14 +499,21 @@ private ModelTransaction getOrCreateResource(String resource, Class<?> type, Mod
...
-            feature = EMFUtil.createEAttribute(service, resource, type, ef -> createEFeatureAnnotation(ef, timestamp));
-            service.getEStructuralFeatures().add(feature);
+            feature = createResource(service, resource, type, timestamp, null, accumulator);
         }
         transaction.addFeature(feature);
         return transaction;
     }
 
+    private EAttribute createResource(EClass service, String resource, Class<?> type, Instant timestamp,
+            Object defaultValue, NotificationAccumulator accumulator) {
```
tcalmant, line 509: Do we need the accumulator here?

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/twin/impl/SensinactDigitalTwinImpl.java
```diff
@@ -262,21 +267,19 @@ private SensinactProviderImpl toProvider(final Provider modelProvider, boolean l
         if (loadServices) {
             // List services
             final List<SensinactService> services = modelProvider.eClass().getEStructuralFeatures().stream()
-                    .map((feature) -> toService(snProvider, (Service) modelProvider.eGet(feature)))
-                    .collect(Collectors.toList());
-            services.add(toService(snProvider, modelProvider.getAdmin()));
+                    .map(feature -> toService(snProvider, feature)).collect(Collectors.toList());
+            services.add(toService(snProvider, modelProvider.eClass().getEStructuralFeature("admin")));
```
tcalmant, line 271: Shouldn't we use `getEStructuralFeature(SensiNactPackage.ADMIN)` ?

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/ModelNexus.java
timothyjward: We will do - the notifications are missing some events at the moment, including linking/unlinking and model change events

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/twin/impl/SensinactDigitalTwinImpl.java
tcalmant, line 271: Shouldn't we use `getEStructuralFeature(SensiNactPackage.ADMIN)` ?

PR #132, opened 2023-02-20
Implemented changes of the new model

juergen-albert:


prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/ModelNexus.java
```diff
@@ -660,36 +648,26 @@ private EReference getServiceForModel(EClass model, String serviceName) {
...
-            if (annotation == null) {
-                // TODO do we have to special case admin, or should it have a metadata
-                // annotation
-                return sensinactPackage.getAdmin().equals(a.getEContainingClass());
-            } else {
-                return annotation.eContents().stream().anyMatch(ModelMetadata.class::isInstance);
-            }
-        });
+    public Stream<ETypedElement> getResourcesForService(EClass svcClass) {
+        List<ETypedElement> result = new ArrayList<>(svcClass.getEAllAttributes());
```
timothyjward, line 652: Is this correct, or should we be filtering for `ResourceAttribute` instances?

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/ModelNexus.java
```diff
@@ -660,36 +648,26 @@ private EReference getServiceForModel(EClass model, String serviceName) {
...
+        result.addAll(svcClass.getEOperations());
```
timothyjward, line 653: Is there a special type of `EOperation` that we should filter for (like there is with`ResourceAttribute`)?

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/ModelNexus.java
```diff
@@ -660,36 +648,26 @@ private EReference getServiceForModel(EClass model, String serviceName) {
...
+        return result.stream();
```
timothyjward, line 654: Stream.concat(svcClass.getEAllAttributes().stream(), svcClass.getEOperations().stream()) would avoid creating an intermediate list.

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/twin/impl/SensinactDigitalTwinImpl.java
```diff
@@ -147,19 +148,23 @@ private SensinactResourceImpl getResource(Provider provider, String model, Strin
...
         }
 
-        final EStructuralFeature svcFeature = provider.eClass().getEStructuralFeature(service);
+        final EReference svcFeature = (EReference) provider.eClass().getEStructuralFeature(service);
         if (svcFeature == null) {
             return null;
         }
 
-        final EStructuralFeature rcFeature = ((EClass) svcFeature.getEType()).getEStructuralFeature(resource);
+        final EClass serviceEClass = (EClass) svcFeature.getEType();
```
timothyjward, line 156:
```suggestion
        final EClass serviceEClass = svcFeature.getEReferenceType();
```

prototype/core/models/src/main/resources/model/sensinact_bak.ecore
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
timothyjward, line 1: Should this file have been checked in?

timothyjward (commented):
A few minor questions/details

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/ModelNexus.java
juergen-albert: Thanks, I know something like this should exists, but I couldn't find it.

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/ModelNexus.java
juergen-albert: I already planed a bit ahead, when I want to work more directly with EObjects using EClasses that don't inherit from our model.

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/ModelNexus.java
juergen-albert: should we sort this concatinated stream automatically sort by name?

timothyjward: We shouldn’t sort it - consumers can sort according to their needs, but in many cases no sorting will be required.

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/ModelNexus.java
timothyjward: Ok - we should add a comment here saying that it may be too permissive and then we should use an issue to plan how we recognise resource attributes in these cases.

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/twin/impl/SensinactDigitalTwinImpl.java
```diff
@@ -147,19 +148,23 @@ private SensinactResourceImpl getResource(Provider provider, String model, Strin
...
-        final EStructuralFeature svcFeature = provider.eClass().getEStructuralFeature(service);
+        final EReference svcFeature = (EReference) provider.eClass().getEStructuralFeature(service);
         if (svcFeature == null) {
             return null;
         }
 
-        final EStructuralFeature rcFeature = ((EClass) svcFeature.getEType()).getEStructuralFeature(resource);
+        final EClass serviceEClass = svcFeature.getEReferenceType();
+
+        final ETypedElement rcFeature = serviceEClass.getEOperations().stream()
```
timothyjward, line 158: Two questions 
1. does this need to be getEAllOperations?
2. Would it make sense to query the structural feature first? Attributes are more common than operations, and the query by name should be faster than searching a list.

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/ModelNexus.java
timothyjward: Also - should this be getEAllOperations()

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/twin/impl/SensinactDigitalTwinImpl.java
juergen-albert: > does this need to be getEAllOperations?
Right now we don't, as we don't have operations on Objects we inherit from. But this makes sense for the Future. Note: If you ever use getEAll<Something>, keep in mind that you will always will get everything from the EObject level as well. I've added a filter  for that.

> Would it make sense to query the structural feature first? Attributes are more common than operations, and the query by name should be faster than searching a list.

Yes, that makes sense.

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/impl/ModelNexus.java
juergen-albert: I've created #134 for this.

juergen-albert:
@timothyjward If you ask me, this is ready to be merged.

timothyjward (approved):

PR #136, opened 2023-02-24
Added @model placeholder in device factory

tcalmant:
* Allows the device factory to set the model of the provider based on a literal or a record entry using `@model`.
* Adds unit tests to the device factory core behaviour
* Fixed wrong type of injected values


prototype/southbound/device-factory/device-factory-core/src/test/java/org/eclipse/sensinact/gateway/southbound/device/factory/impl/BulkDTOPromise.java
```diff
@@ -0,0 +1,117 @@
+/*********************************************************************
```
timothyjward, line 1: Is there a reason not to just use Promises.resolved(dto) and remove this type altogether?

tcalmant: I reused some previous code. I'll change it in the 3 device factory projects.

PR #141, opened 2023-03-02
Provider push implementation

juergen-albert:
* Complete Provider instances can be pushed
* Notifications are now handled via EMF Change Adapters


juergen-albert:
@timothyjward I know this is a bigger one. While implementing this, I found a few notifications that have been missing or some that seemed a bit wrong to me. Before I do long writeup, I'd be happy to have a quick meeting to talk about it.

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/ModelNexus.java
```diff
@@ -240,17 +246,23 @@ public void linkProviders(String parentProvider, String childProvider, Instant t
 
         NotificationAccumulator accumulator = notificationAccumulator.get();
 
-        Provider parent = providers.get(parentProvider);
+        EObject parent = providers.get(parentProvider);
```
timothyjward, line 249: Isn't providers typed to be a Provider?

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/ModelNexus.java
```diff
@@ -240,17 +246,23 @@ public void linkProviders(String parentProvider, String childProvider, Instant t
...
 
-        Provider child = providers.get(childProvider);
+        EObject child = providers.get(childProvider);
 
         if (parent == null) {
             throw new IllegalArgumentException("No parent provider " + parentProvider);
+        } else if (!Provider.class.isInstance(parent)) {
```
timothyjward, line 255: I don't think that this can happen due to the generics on the providers map

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/ModelNexus.java
```diff
@@ -272,19 +284,23 @@ public void unlinkProviders(String parentProvider, String childProvider, Instant
...
+        EObject parent = providers.get(parentProvider);
 
-        Provider parent = providers.get(parentProvider);
-
-        Provider child = providers.get(childProvider);
+        EObject child = providers.get(childProvider);
 
         if (parent == null) {
             throw new IllegalArgumentException("No parent provider " + parentProvider);
+        } else if (!Provider.class.isInstance(parent)) {
```
timothyjward, line 293: As above

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/ModelNexus.java
```diff
@@ -293,60 +309,38 @@ public void unlinkProviders(String parentProvider, String childProvider, Instant
     public void handleDataUpdate(String modelName, Provider provider, EStructuralFeature serviceFeature,
             EStructuralFeature resourceFeature, Object data, Instant timestamp) {
 
-        Instant metaTimestamp = timestamp == null ? Instant.now() : timestamp;
+        try (Transaction t = Transaction.startTransaction()) {
```
timothyjward, line 312: What does this do? I can't see any reason for it to exist.

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/ModelNexus.java
```diff
@@ -358,43 +352,63 @@ public void handleDataUpdate(String modelName, Provider provider, EStructuralFea
...
      */
+    private Provider doCreateProvider(String modelName, EClass model, String providerName, Instant timestamp) {
+        return doCreateProvider(modelName, model, providerName, timestamp, true);
+    }
+
     private Provider doCreateProvider(String modelName, EClass model, String providerName, Instant timestamp,
-            NotificationAccumulator accumulator) {
+            boolean createAdmin) {
+
+        try (Transaction t = Transaction.startTransaction()) {
```
timothyjward, line 362: Another transaction that is never used

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/ModelNexus.java
```diff
@@ -403,37 +417,47 @@ public Provider createProviderInstance(String modelName, String providerName, In
...
-            provider = doCreateProvider(modelName, getMandatoryModel(modelName), providerName, timestamp, accumulator);
+            provider = doCreateProvider(modelName, getMandatoryModel(modelName), providerName, timestamp);
         }
 
         return provider;
     }
 
     public Provider getProvider(String providerName) {
-        return providers.get(providerName);
+        return transformToProvider(providers.get(providerName));
```
timothyjward, line 427: Not sure why this is here

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/emf/change/MetadataChangeAdapter.java
```diff
@@ -0,0 +1,97 @@
...
+    public Instant getPreviousTimestamp() {
+        return previousTimestamp;
+    }
+
+    private Metadata getMetaData() {
+        return (Metadata) getTarget();
+    }
+
+    private void notifyUpdate(Notification msg, NotificationAccumulator accumulator) {
+//        Metadata metadata = getMetaData();
```
timothyjward, line 72: This shouldn't be commended out code

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/model/nexus/emf/change/Transaction.java
```diff
@@ -0,0 +1,59 @@
+/*********************************************************************
```
timothyjward, line 1: I think this type can be deleted

prototype/core/impl/src/test/java/org/eclipse/sensinact/prototype/model/nexus/impl/NexusTest.java
```diff
@@ -110,14 +120,101 @@ public class BasicModelTests {
         @Test
         void basicTest() {
 
+            resourceSet.setURIConverter(new URIConverter() {
```
timothyjward, line 123: Could we just mock this?

prototype/pom.xml
```diff
@@ -296,11 +300,31 @@
...
+        <groupId>org.eclipse.emf</groupId>
+        <artifactId>org.eclipse.emf.ecore.change</artifactId>
+        <version>${emf.compare.change.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.eclipse.emf</groupId>
+        <artifactId>org.eclipse.emf.compare</artifactId>
+        <version>${emf.compare.version}</version>
+      </dependency>
+      <dependency>
```
timothyjward, line 313: Is there any way that we can avoid guava? It's a versioning nightmare as they always bump the major version.

juergen-albert: Not for now, as this is a dependency of emf compare. As they build with tycho, their POMs don't have a dependency section. Thus we need to add their transient dependencies. 

When I find the time, I want to create PR for them to replace Guava, as this seems to be a historic dependency from the times before Function, Supplier, Lists and so on. So, it can be completely replaced with standard Java means.

PR #154, opened 2023-05-10
add grouped Mode

vwfox:
add group centered view
add infobox
add docker/docker-compose/docker-build

add licence headers

lint editorconfig

fix whitspace


stbischof:
the jena logo svg#s might not live in sensinact

replaye YYYY with 2023

* Copyright (c) YYYY Contributors to the Eclipse Foundation.

* Copyright (c) 2023 Contributors to the Eclipse Foundation.

stbischof:
@timothyjward any opinions or could i merge?

timothyjward:
> @timothyjward any opinions or could i merge?

Is there any way to download the icons in `prototype/webclient/src/assets/icons/` as part of the build or npm dependency gathering rather than having them in the repository? The same goes for the fonts in `prototype/webclient/src/assets/fonts` which are also missing a `NOTICE` file to say where they came from. 

All of the assets in these folders appear to be available as downloadable artefacts from their creators, so it would be nice not to fill this repository with them and to further emphasise that they're an external dependency. The background images appear to be specially created rather than downloaded, so it makes sense for them to stay despite their size.

Finally, I'm not seeing a package-lock.json - don't we want to check this file in to get consistent builds?

PR #167, opened 2023-06-14
Implementation of pull-based resources

tcalmant:
* Uses `isExternalGet` and `isExternalSet` flags of the `ResourceAttribute` to check if an external handler must be called
* Adds the `externalGetCacheMs` entry in `ResourceAttribute` to define the time to wait before calling again the external Get handler
* The value returned by the handlers is stored in the twin
* Added `GetLevel` to define 3 levels of resource access:
  * `WEAK`: always return the twin value, never call the external get
  * `NORMAL`: call the external get if no twin value is available or if the `externalGetCacheMs` duration expired
  * `STRONG`: always call the external get.
* Doesn't work with resources defined outside the ModelBuilder, like pre-defined admin resources.

Also adds a small fix in the history provider integration tests to protect against a null pointer when Docker is not available.

### DIFF
--- a/prototype/core/annotation/src/main/java/org/eclipse/sensinact/core/annotation/verb/GET.java
+++ b/prototype/core/annotation/src/main/java/org/eclipse/sensinact/core/annotation/verb/GET.java
@@ -17,6 +17,8 @@ import java.lang.annotation.Repeatable;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+import java.time.temporal.ChronoUnit;
+import java.util.concurrent.TimeUnit;
 
 import org.eclipse.sensinact.core.annotation.dto.NullAction;
 
@@ -72,6 +74,19 @@ public @interface GET {
      * @return
      */
     Class<?> type() default Object.class;
+
+    /**
+     * Duration of the value cache. If the value is requested using a CACHED get and
+     * the cache duration is exceeded, the annotated method will be called.
+     *
+     * The default duration is in {@link TimeUnit#MILLISECONDS}.
+     */
+    long cacheDuration() default 500;
+
+    /**
+     * Unit of the value cache duration. Defaults to {@link ChronoUnit#MILLIS}.
+     */
+    ChronoUnit cacheDurationUnit() default ChronoUnit.MILLIS;
 
     NullAction onNull() default NullAction.IGNORE;
 
--- a/prototype/core/annotation/src/main/java/org/eclipse/sensinact/core/annotation/verb/GET.java
+++ b/prototype/core/annotation/src/main/java/org/eclipse/sensinact/core/annotation/verb/GET.java
@@ -18,7 +18,6 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.time.temporal.ChronoUnit;
-import java.util.concurrent.TimeUnit;
 
 import org.eclipse.sensinact.core.annotation.dto.NullAction;
 
@@ -79,7 +78,7 @@ public @interface GET {
      * Duration of the value cache. If the value is requested using a CACHED get and
      * the cache duration is exceeded, the annotated method will be called.
      *
-     * The default duration is in {@link TimeUnit#MILLISECONDS}.
+     * The default duration is in {@link ChronoUnit#MILLIS}.
      */
     long cacheDuration() default 500;
 
--- a/prototype/core/api/src/main/java/org/eclipse/sensinact/core/command/GetLevel.java
+++ b/prototype/core/api/src/main/java/org/eclipse/sensinact/core/command/GetLevel.java
@@ -1,0 +1,38 @@
+/*********************************************************************
+* Copyright (c) 2023 Contributors to the Eclipse Foundation.
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which is available at https://www.eclipse.org/legal/epl-2.0/
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*   Kentyou - initial implementation
+**********************************************************************/
+package org.eclipse.sensinact.core.command;
+
+/**
+ * Variants of the GET command
+ */
+public enum GetLevel {
+
+    /**
+     * Default level. For pull-based values, if the cached value is older than the
+     * threshold, pull the real value, else return the cached one. For push-based
+     * values, returns the last cached value.
+     */
+    CACHED,
+
+    /**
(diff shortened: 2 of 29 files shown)


prototype/core/annotation/src/main/java/org/eclipse/sensinact/core/annotation/verb/GET.java
```diff
@@ -73,6 +75,19 @@
...
+     * the cache duration is exceeded, the annotated method will be called.
+     *
+     * The default duration is in {@link TimeUnit#MILLISECONDS}.
+     */
+    long cacheDuration() default 500;
+
+    /**
+     * Unit of the value cache duration. Defaults to {@link ChronoUnit#MILLIS}.
+     */
+    ChronoUnit cacheDurationUnit() default ChronoUnit.MILLIS;
```
timothyjward, line 89: Should this be TimeUnit rather than ChronoUnit? My reasoning is that we're dealing with relative times not dates, so TimeUnit is more appropriate. It also links with the JavaDoc for `cacheDuration` which says `TimeUnit`.

prototype/core/api/src/main/java/org/eclipse/sensinact/core/command/GetLevel.java
```diff
@@ -0,0 +1,38 @@
...
+     * threshold, pull the real value, else return the cached one. For push-based
+     * values, returns the last cached value.
+     */
+    CACHED,
+
+    /**
+     * Weak Get: always return the last value in cache. Don't try to pull a real
+     * value
+     */
+    WEAK,
```
timothyjward, line 31: We should document the behaviour if the value has never been set. Would it do a pull in this case? I think probably not, but we should be explicit.

prototype/core/api/src/main/java/org/eclipse/sensinact/core/command/GetLevel.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Hard Get: always ask for the real value and put it in cache. Acts like
+     * {@link #CACHED} for pushed values.
+     */
+    HARD,
```
timothyjward, line 37: This is notionally the opposite of a `WEAK` get so I think that this constant name should be `STRONG`, or that `WEAK` should be changed to `SOFT`?

prototype/core/api/src/main/java/org/eclipse/sensinact/core/model/ResourceBuilder.java
```diff
@@ -85,20 +85,21 @@
...
 
     /**
-     * Set a getter function to be called
-     *
-     * @param getter
-     * @return
+     * This resource has a dynamic get behavior. Can't work on resource with an action.
      */
-    ResourceBuilder<B, T> withGetter(Supplier<T> getter);
+    ResourceBuilder<B, T> withGetter();
```
timothyjward, line 90: Should we have a boolean version as well so that people can say `withGetter(false)` when copying another model?

prototype/core/api/src/main/java/org/eclipse/sensinact/core/model/ResourceBuilder.java
```diff
@@ -85,20 +85,21 @@
...
-     * @return
+     * @param cacheDuration Duration of the cache
+     */
+    ResourceBuilder<B, T> withGetterCache(Duration cacheDuration);
+
+    /**
+     * This resource has a dynamic set behavior. Can't work on resource with an action.
      */
-    ResourceBuilder<B, T> withSetter(Consumer<T> setter);
+    ResourceBuilder<B, T> withSetter();
```
timothyjward, line 102: As with `withGetter` a boolean may be useful?

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/command/impl/GatewayThreadImpl.java
```diff
@@ -153,6 +155,41 @@ private Promise<Object> performAction(String model, String provider, String serv
         return d.getPromise();
     }
 
+    private <T> Promise<TimedValue<T>> pullResourceValue(String model, String provider, String service,
```
timothyjward, line 158: This behaviour should live outside the gateway thread, probably in the model nexus or the resource itself. 
Having reflected, we shouldn't be grabbing the gateway thread's promise factory for the work either as that's really for command use. GET/SET/ACT should share their own thread pool for doing the calls and then update the twin on the gateway thread with a command.

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/command/impl/GatewayThreadImpl.java
```diff
@@ -153,6 +155,41 @@ private Promise<Object> performAction(String model, String provider, String serv
...
+                    d.fail(e);
+                }
+            });
+        } catch (Exception e) {
+            d.fail(e);
+        }
+        return d.getPromise();
+    }
+
+    private <T> Promise<TimedValue<T>> pushResourceValue(String model, String provider, String service,
```
timothyjward, line 175: As with pull, this shouldn't be here. It should be done in the nexus/resource

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/impl/SensiNactSessionImpl.java
```diff
@@ -216,8 +216,31 @@ private <T> T safeGetValue(Promise<T> promise) {
...
+    public <T> TimedValue<T> getResourceTimedValue(String provider, String service, String resource, Class<T> clazz) {
+        return safeExecute(new AbstractTwinCommand<TimedValue<T>>() {
+            @Override
+            protected Promise<TimedValue<T>> call(SensinactDigitalTwin model, PromiseFactory pf) {
+                final SensinactResource sensinactResource = model.getResource(provider, service, resource);
+                if (sensinactResource != null) {
+                    if (sensinactResource.getResourceType() == ResourceType.ACTION) {
+                        return pf.resolved(null);
+                    }
+                    // TODO: get the GetLevel from argument
```
timothyjward, line 237: Does this need tidying up?

prototype/core/annotation/src/main/java/org/eclipse/sensinact/core/annotation/verb/GET.java
tcalmant: As we use `Duration` internally, I'd prefer to keep using `ChronoUnit` and update the Javadoc

prototype/core/api/src/main/java/org/eclipse/sensinact/core/command/GetLevel.java
tcalmant: As discussed, we'll use `WEAK`, `NORMAL` and `STRONG`

PR #173, opened 2023-06-28
Added utility methods to GeoJSON

tcalmant:
* Adds support for hashCode(), equals() and toString() to GeoJSON types
* Adds a GeoJsonUtils class with static utility methods to create GeoJSON objects

### DIFF
--- a/prototype/core/geo-json/src/main/java/org/eclipse/sensinact/gateway/geojson/Coordinates.java
+++ b/prototype/core/geo-json/src/main/java/org/eclipse/sensinact/gateway/geojson/Coordinates.java
@@ -12,6 +12,8 @@
 **********************************************************************/
 package org.eclipse.sensinact.gateway.geojson;
 
+import java.util.Objects;
+
 import org.eclipse.sensinact.gateway.geojson.internal.CoordinatesDeserializer;
 import org.eclipse.sensinact.gateway.geojson.internal.CoordinatesSerializer;
 
@@ -36,4 +38,45 @@ public class Coordinates {
      * The elevation will be {@link Double#NaN} if not set
      */
     public double elevation = Double.NaN;
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(longitude, latitude, elevation);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj instanceof Coordinates) {
+            // Chosen according to https://xkcd.com/2170/
+            final double epsilon = 0.0000001d;
+            final Coordinates other = (Coordinates) obj;
+            if (Math.abs(other.longitude - longitude) >= epsilon) {
+                return false;
+            }
+
+            if (Math.abs(other.latitude - latitude) >= epsilon) {
+                return false;
+            }
+
+            if (Double.isFinite(other.elevation) && Double.isFinite(elevation)) {
+                // Check elevations if both are finite
+                return Math.abs(other.latitude - latitude) < epsilon;
+            }
+
+            // Equality if both elevations are not finite
+            return !Double.isFinite(other.elevation) && !Double.isFinite(elevation);
+        }
+
+        return false;
+    }
+
+    @Override
+    public String toString() {
+        String repr = "(lon=" + longitude + ",lat=" + latitude;
+        if (Double.isFinite(elevation)) {
+            // Add elevation
+            repr += ",alt=" + elevation;
+        }
+        return repr + ")";
+    }
 }
--- a/prototype/core/geo-json/src/main/java/org/eclipse/sensinact/gateway/geojson/Coordinates.java
+++ b/prototype/core/geo-json/src/main/java/org/eclipse/sensinact/gateway/geojson/Coordinates.java
@@ -46,7 +46,7 @@ public class Coordinates {
 
     @Override
     public boolean equals(Object obj) {
-        if (obj instanceof Coordinates) {
+        if (obj != null && obj.getClass() == Coordinates.class) {
             // Chosen according to https://xkcd.com/2170/
             final double epsilon = 0.0000001d;
             final Coordinates other = (Coordinates) obj;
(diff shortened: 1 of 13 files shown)


prototype/core/geo-json/src/main/java/org/eclipse/sensinact/gateway/geojson/Coordinates.java
```diff
@@ -36,4 +38,45 @@ public class Coordinates {
...
     public double elevation = Double.NaN;
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(longitude, latitude, elevation);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj instanceof Coordinates) {
```
timothyjward, line 49: This check isn't considered best practice for `equals` unless the type is `final`. If someone subclasses Coordinates then this test will be true, even though we're doing a potentially incomplete check.

```suggestion
        if (obj != null && obj.getClass() == Coordinates.class) {
```

prototype/core/geo-json/src/main/java/org/eclipse/sensinact/gateway/geojson/Coordinates.java
```diff
@@ -36,4 +38,45 @@ public class Coordinates {
...
+            // Equality if both elevations are not finite
+            return !Double.isFinite(other.elevation) && !Double.isFinite(elevation);
+        }
+
+        return false;
+    }
+
+    @Override
+    public String toString() {
+        String repr = "(lon=" + longitude + ",lat=" + latitude;
```
timothyjward, line 75: A StringBuilder would be faster and more readable code.

prototype/core/geo-json/src/main/java/org/eclipse/sensinact/gateway/geojson/Feature.java
```diff
@@ -27,4 +28,23 @@ public Feature() {
...
+    @Override
+    public int hashCode() {
+        return Objects.hash(super.hashCode(), id, geometry, properties);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (super.equals(obj)) {
+            final Feature other = (Feature) obj;
+            return Objects.equals(id, other.id) && Objects.equals(geometry, ((Feature) obj).geometry)
```
timothyjward, line 40:
```suggestion
            return Objects.equals(id, other.id) && Objects.equals(geometry, other.geometry)
```

prototype/core/geo-json/src/main/java/org/eclipse/sensinact/gateway/geojson/GeoJsonObject.java
```diff
@@ -61,4 +63,44 @@ protected GeoJsonObject(GeoJsonType type) {
...
 
     public List<Double> bbox;
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(type, bbox);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
timothyjward, line 73: I don't think this should be an override of `equals` as it's not a complete equals method. 

If it were:

    protected boolean checkParentEquals(Object obj) {

Then that would prevent any confusion that this might be usable on its own.

prototype/core/geo-json/src/main/java/org/eclipse/sensinact/gateway/geojson/GeoJsonObject.java
```diff
@@ -61,4 +63,44 @@ protected GeoJsonObject(GeoJsonType type) {
...
+        if (obj instanceof GeoJsonObject) {
```
timothyjward, line 74: Add a comment indicating that this is safe because it's only intended to be used as part of an equals in a concrete subclass.

prototype/core/geo-json/src/main/java/org/eclipse/sensinact/gateway/geojson/GeoJsonObject.java
```diff
@@ -61,4 +63,44 @@ protected GeoJsonObject(GeoJsonType type) {
...
+                    && Objects.equals(foreignMembers, other.foreignMembers);
+        }
+
+        return false;
+    }
+
+    /**
+     * Returns the object description
+     */
+    protected abstract String getObjectDescription();
```
timothyjward, line 86:
```suggestion
    protected abstract String getObjectDescription(StringBuilder builder);
```

Passing a StringBuilder here would allow us to avoid creating lots of intermediate Strings.

prototype/core/geo-json/src/main/java/org/eclipse/sensinact/gateway/geojson/GeoJsonObject.java
```diff
@@ -61,4 +63,44 @@ protected GeoJsonObject(GeoJsonType type) {
...
+
+    @Override
+    public String toString() {
+        final List<String> parts = new ArrayList<>();
```
timothyjward, line 90: If we start with a StringBuilder we can pass it to the subclass method to populate.

Also, shouldn't we start by putting the `type` into the list/StringBuilder so we know what type of Object it is?

prototype/core/geo-json/src/main/java/org/eclipse/sensinact/gateway/geojson/GeometryCollection.java
```diff
@@ -27,4 +28,21 @@ public GeometryCollection() {
...
     public List<Geometry> geometries;
 
+    @Override
+    public int hashCode() {
+        return Objects.hash(super.hashCode(), geometries);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj instanceof GeometryCollection) {
```
timothyjward, line 38: Shouldn't this call the superclass method?

timothyjward (changes requested):
There are a couple of small but significant things that need fixing for `equals`. `toString` is fine, but could be made faster and more readable by passing around a `StringBuilder`.

timothyjward (approved):

PR #175, opened 2023-07-10
Add a metrics service for SensiNact

tcalmant:
This PR adds an `IMetricsManager` service and types that hide the use of [Metrics](https://metrics.dropwizard.io/).
Some metrics have been added to the gateway thread, the whiteboard ACT/GET/SET methods execution times and the number of sessions.
When enabled, metrics can be visible in the standard output of sensiNact and/or through a sensiNact provider.
Documentation have been added to indicate how to add metrics.


prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/command/impl/GatewayThreadImpl.java
```diff
@@ -155,7 +164,15 @@ public void run() {
...
                 currentItem.set(item);
-                item.doWork();
+
+                metrics.getCounter("sensinact.tasks.pending").dec();
+                metrics.getHistogram("sensinact.tasks.pending.hist").update(work.size());
+                try (IMetricTimer timer = metrics.withTimer("sensinact.task.time")) {
+                    item.doWork();
+                } catch (Exception e) {
+                    // Ignore: the timer shouldn't fail on close
+                    e.printStackTrace();
```
timothyjward, line 174: This will break the behaviour where `item.doWork()` throws an `InterruptedException`

The Timer (and other interfaces) should override the close method to remove the exception as described in https://stackoverflow.com/questions/25959779/why-close-method-of-java-lang-autocloseable-throws-exception-but-close-method-o#answer-49515309 - this will allow try-with-resources, but it will never need a `catch (Exception)` block

prototype/core/api/src/main/java/org/eclipse/sensinact/core/metrics/IMetricTimer.java
```diff
@@ -0,0 +1,25 @@
...
+ * A simple timer that holds the time taken by a task.
+ *
+ * Timers are started immediately when {@link IMetricsManager#withTimer(String)}
+ * returns and run until their {@link #close()} method is called.
+ *
+ * Timers are {@link AutoCloseable} and can be used in a try-with-resources
+ * pattern.
+ */
+public interface IMetricTimer extends INamedMetric, AutoCloseable {
+}
```
timothyjward, line 25: Add

    /**
     * Override to remove the exception
     */
    @Override
    void close();

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/impl/SessionManager.java
```diff
@@ -50,6 +56,16 @@ public class SessionManager implements SensiNactSessionManager, TypedEventHandle
...
     private final Map<String, String> userDefaultSessionIds = new HashMap<>();
 
+    @Activate
+    void activate() {
+        metrics.registerGauge("sensinact.sessions", () -> sessions.size());
+    }
+
+    @Deactivate
+    void deactivate() {
+        metrics.unregisterGauge("sensinact.sessions");
```
timothyjward, line 66: The listener pattern (which this is an example of) can be a challenge in OSGi. Could this be turned into a whiteboard?

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/metrics/impl/MetricsManager.java
```diff
@@ -0,0 +1,332 @@
...
+                }
+            }
+        }
+    }
+
+    @Override
+    public IMetricCounter getCounter(String name) {
+        if (!isEnabled(name)) {
+            // Return a dummy counter
+            return new DummyCounter(name);
```
timothyjward, line 317: What happens if the configuration is updated to enable it after the counter is obtained? Shouldn't the Counter implementation stay the same but just ignore the calls when it is disabled?

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/metrics/impl/MetricsManager.java
```diff
@@ -0,0 +1,332 @@
...
+        }
+
+        return new MetricsCounter(registry, name);
+    }
+
+    @Override
+    public IMetricsHistogram getHistogram(String name) {
+        if (!isEnabled(name)) {
+            // Return a dummy counter
+            return new DummyHistogram(name);
```
timothyjward, line 327: As with a counter, this feels like it should be a smarter implementation that can be enabled/disabled.

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/metrics/impl/MetricsManager.java
```diff
@@ -0,0 +1,332 @@
...
+    public void disableMetrics(String... names) {
+        if (names != null) {
+            activeMetrics.removeAll(Arrays.asList(names));
+        }
+    }
+
+    @Override
+    public IMetricTimer withTimer(String name) {
+        if (!isEnabled(name)) {
+            return new DummyTimer(name);
```
timothyjward, line 272: Note that timer is different from the counter/histogram comments below. This is being obtained for every call so a dummy object works well.

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/whiteboard/impl/SensinactWhiteboard.java
```diff
@@ -508,7 +516,7 @@ public Promise<Object> act(String model, String provider, String service, String
         } else {
             Deferred<Object> d = promiseFactory.deferred();
             promiseFactory.executor().execute(() -> {
-                try {
+                try (IMetricTimer timer = metrics.withTimer("sensinact.whiteboard.act")) {
```
timothyjward, line 519: I'm not sure that this is the correct way to measure this execution time. By measuring this way we lose the time between the call to `resolveWith` and the actual resolution if the action returns a promise.

prototype/core/impl/src/main/java/org/eclipse/sensinact/prototype/whiteboard/impl/SensinactWhiteboard.java
```diff
@@ -606,7 +614,7 @@ public <T> Promise<TimedValue<T>> pushValue(String model, String provider, Strin
             final Deferred<TimedValue<T>> d = promiseFactory.deferred();
             final SetMethod setMethod = opt.get();
             promiseFactory.executor().execute(() -> {
-                try {
+                try (IMetricTimer timer = metrics.withTimer("sensinact.whiteboard.push")) {
```
timothyjward, line 617: See above

prototype/docs/core/metrics.md
```diff
@@ -0,0 +1,290 @@
...
+The method expects a consumer of `BulkGenericDto` a object that will be called after each report generation.
+The registered listener has an ID that must be given to the `unregisterListener` method.
+Each listener must be unregistered explicitly.
+
+The `BulkGenericDto` provides a list of `GenericDto` objects that represents the resources as defined the [Core metrics section](#sensinact-metrics) and in the [metrics provider description](#note-on-the-metrics-provider).
+
+## Store metrics
+
+It is possible to store your own metrics in sensiNact by calling the `IMetricsManager` service.
+Make sure that the gauges you register are quick enough to note slow the metrics report generation.
```
timothyjward, line 110:
```suggestion
Make sure that the callback functions for any gauges you register execute quickly. If they block, or require significant computation, then they will significantly slow the metrics report generation *and* impact the measured metrics.
```

prototype/docs/core/metrics.md
```diff
@@ -0,0 +1,290 @@
...
+    executeTask();
+}
+```
+
+Unlike other statistics, you cannot reuse a timer instance.
+
+### Use other metrics
+
+Counters and histograms can be used inline.
+It is not recommended to keep it as a class member as the returned instance depends on the metrics activation.
```
timothyjward, line 271: I think that this should be fixed in the implementation of the metrics, otherwise it just makes the usage harder.

timothyjward (changes requested):
Some minor comments, but the `IMetricTimer` change (adding an overridden close without an exception) is an important usability improvement.

timothyjward (approved):

PR #192, opened 2023-07-19
Support for multiple repository and features folders in launcher configuration

tcalmant:
This PR adds the possibility to define multiple repository and feature folders in the launcher configuration.
Previous format, *i.e.* single folder definition, is still supported.

The search for a bundle or a feature file is done in the order of definition of their folders in the configuration.

This allows for example to have a read-only repository for the sensiNact core bundles repository and to have multiple repositories for instance-specific features.


distribution/launcher/src/main/java/org/eclipse/sensinact/gateway/launcher/FeatureLauncher.java
```diff
@@ -296,7 +319,13 @@ private void addOrUpdate(String feature, List<String> installProgress) throws Co
         // Start
         for (Bundle b : installed) {
             try {
-                b.start();
+                BundleRevision rev = b.adapt(BundleRevision.class);
+                if (rev == null || (rev.getTypes() & BundleRevision.TYPE_FRAGMENT) == 0) {
```
timothyjward, line 323:
```suggestion
                if (rev != null && (rev.getTypes() & BundleRevision.TYPE_FRAGMENT) == 0) {
```
The revision will only be null for UNINSTALLED bundles which we should also not try to start!

distribution/launcher/src/main/java/org/eclipse/sensinact/gateway/launcher/FeatureLauncher.java
```diff
@@ -307,19 +336,29 @@ private Feature loadFeature(String feature) {
...
         } else {
             String simpleFileName = feature;
             if (!simpleFileName.endsWith(".json")) {
                 simpleFileName += ".json";
             }
 
-            featureFile = featureDir.resolve(simpleFileName);
+            for (Path featureDir : featureDirs) {
+                Path testedFile = featureDir.resolve(simpleFileName);
+                if (Files.exists(testedFile)) {
```
timothyjward, line 354: Should this be 
```suggestion
                if (Files.isRegularFile(testedFile)) {
```

distribution/launcher/src/main/java/org/eclipse/sensinact/gateway/launcher/FeatureLauncher.java
```diff
@@ -307,19 +336,29 @@ private Feature loadFeature(String feature) {
...
+                    featureFile = testedFile;
+                    break;
+                }
+            }
         }
 
-        if (Files.isRegularFile(featureFile)) {
+        if (featureFile != null && Files.isRegularFile(featureFile)) {
```
timothyjward, line 361: If we add the isRegularFile check above then this can be
```suggestion
        if (featureFile != null) {
```

timothyjward (approved):

distribution/launcher/src/main/java/org/eclipse/sensinact/gateway/launcher/FeatureLauncher.java
tcalmant: The `getFileFromRepository` method can return an invalid path, so the isRegularFile check should stay.

PR #201, opened 2023-08-02
Fix cases where a northbound result had 2 'type' keys

tcalmant:
In some cases, the JSON result sent by sensiNact had two "type" keys at root level.
This PR fixes the Jackson serialization annotations and adds a test to avoid reproducing the error.


timothyjward:
The code changes look good, but some of the REST tests are now failing with the wrong `type` set in responses.

tcalmant:
I found the issue, I'm preparing the commit

PR #218, opened 2023-08-30
Documentation improvements

timothyjward:
Add some worked examples for the gateway

### DIFF
--- a/docs/examples/Interacting.md
+++ b/docs/examples/Interacting.md
@@ -234,3 +234,6 @@ A successful response will look like the following:
 
 After a `SET` subsequent `GET` requests will return the previously set value.
 
+> [!NOTE]
+> If your resource is `FIXED` or `UPDATABLE` then it cannot be set by the northbound API. Also, the behaviour of `SET` requests may be slightly different if your resource is [pull-based](../southbound/custom/custom.md#SET-Methods) and the device returns a different value.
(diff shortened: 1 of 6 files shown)


docs/examples/Interacting.md
```diff
@@ -0,0 +1,236 @@
...
+  "response": {
+    "name": "friendlyName",
+    "timestamp": 1693407687936,
+    "type": "java.lang.String",
+    "value":"Lassie"
+  }
+}
+```
+
+After a `SET` subsequent `GET` requests will return the previously set value.
```
tcalmant, line 235: Should we add a warning about pull/push-based resources which can have a different behavior?

tcalmant (approved):

docs/examples/Interacting.md
timothyjward:
```suggestion
After a `SET` subsequent `GET` requests will return the previously set value.

> [!NOTE]
> If your resource is `FIXED` or `UPDATABLE` then it cannot be set by the northbound API. Also, the behaviour of `SET` requests may be slightly different if your resource is [pull-based](../southbound/custom/custom.md#GET-Methods) and the device returns a different value.
```

tcalmant: Looks good to me

PR #233, opened 2023-09-15
Added support for MQTT authentication

tcalmant:
Added support for basic MQTT authentication (username/password) to the MQTT client.
Also updated the documentation.


timothyjward (commented):
This looks good, but we should create an issue to add configuration for a trust store/pinned certificate (rather than relying on default trust), and possibly support client auth certificates

tcalmant:
> This looks good, but we should create an issue to add configuration for a trust store/pinned certificate (rather than relying on default trust), and possibly support client auth certificates

Agreed.

timothyjward (approved):

PR #239, opened 2023-09-21
Add a figure describing the data model

timothyjward:


tcalmant (changes requested):
The CSS styling should be added as layout customization instead of being applied inside the Markdown files (core/CoreModel.md and index.md).

This can be done by:
1. Creating the file docs/_static/css/style.css with the custom rules
2. Creating the file docs/_templates/layout.html with the following content:
   ```jinja2
   {# Extend layout with a custom CSS file #}
   {% extends "!layout.html" %}

   {% set css_files = css_files + ["_static/css/style.css"] %}
   ```

tcalmant (approved):

PR #297, opened 2024-01-02
[BREAKING] Device Factory enhancements

tcalmant:
This PR provides two changes decided after discussions with @juergen-albert about a better integration of the Device Factory with EMF:
* **BREAKING:** Model/Provider/Service/Resource names now follow the Java identifiers specification
  * As names are integrated into EMF, they must be valid Java identifiers
  * Invalid characters are now replaced by an underscore (`_`) instead of a hyphen (`-`)
  * A new mapping option has been added: `names.ascii`
    * If true, names are ensured to be only ASCII letters, digits or underscores. Diacritics are normalized first.
    * If false (*default*), names are Java identifier characters (see [`Character.isJavaIdentifierPart`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Character.html#isJavaIdentifierPart(char))) or underscores
  * If a name is a Java reserved keyword or if it starts with a digit, it is then prefixed by an underscore.
* *NEW*: Added a `model.raw` mapping option, that disables the normalization of the `@model` content. This allows to use URLs as models.

*Note:* this is currently a draft PR, the documentation has to be updated

### DIFF
--- a/docs/source/getting-started.md
+++ b/docs/source/getting-started.md
@@ -332,8 +332,8 @@ Here is the configuration of the HTTP device factory, where we declare a periodi
     "utc_offset_seconds": 0
   }
   ```
-* maps the response JSON as follows:
-   * the result is associated to a fixed provider name: `grenoble-weather`. `@Provider` is a special placeholder to name the provider that holds the data we received, this is the only mandatory field in a mapping definition.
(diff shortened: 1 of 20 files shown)


tcalmant:
Closed as not necessary since #324

PR #312, opened 2024-01-16
Ensure that the MQTT Client connects to servers eventually

timothyjward:
If the server is not available then we should use the reconnection delay to wait and then reconnect.

### DIFF
--- a/southbound/mqtt/mqtt-client/src/main/java/org/eclipse/sensinact/gateway/southbound/mqtt/impl/MqttClientHandler.java
+++ b/southbound/mqtt/mqtt-client/src/main/java/org/eclipse/sensinact/gateway/southbound/mqtt/impl/MqttClientHandler.java
@@ -12,8 +12,10 @@
 **********************************************************************/
 package org.eclipse.sensinact.gateway.southbound.mqtt.impl;
 
+import java.net.ConnectException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.time.Duration;
 import java.util.Collections;
 import java.util.IdentityHashMap;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


southbound/mqtt/mqtt-client/src/main/java/org/eclipse/sensinact/gateway/southbound/mqtt/impl/MqttClientHandler.java
```diff
@@ -230,11 +254,18 @@ public void connectionLost(final Throwable cause) {
             @Override
             public void run() {
                 try {
-                    client.reconnect();
+                    client.connect(connectOptions);
+                } catch (MqttSecurityException se) {
```
tcalmant, line 258: Shouldn't we log the security exception?

southbound/mqtt/mqtt-client/src/main/java/org/eclipse/sensinact/gateway/southbound/mqtt/impl/MqttClientHandler.java
```diff
@@ -115,24 +124,39 @@ public void activate(final MqttClientConfiguration config) throws Exception {
...
+            client.connect(connectOptions);
+        } catch (MqttException e) {
+            if(e.getCause() instanceof ConnectException) {
+                connectionLost(e);
+            } else {
+                throw e;
+            }
+        }
+
+        subscribe();
```
tcalmant, line 144: We should call subscribe and log the "MQTT client started" trace inside the try block: if the connect fails, we shouldn't try to subscribe

tcalmant (changes requested):
Tests are OK
I have a small doubt on calling subscribe() even if the connection fails in validate()
Also, the editorconfig check is complaining about trailing whitespaces

PR #330, opened 2024-01-29
Bump slf4j to v2.0.11 (Issue #328)

dleangen:
Updating to a more recent version of slf4j, as the current version is an older version.


dleangen:
I also removed references to the log4j-over-slf4j bundle, as it is not being used anywhere.

Issue #331, opened 2024-01-29
Remove dependency on log4j-over-slf4j


dleangen:
This dependency no longer appears to be needed. Simply removing it causes no apparent issues.

dleangen:
Perhaps this is not "important" enough to create an issue. Closing.

PR #332, opened 2024-01-29
(#331) Remove dependency on log4j-over-slf4j

dleangen:


dleangen:
Investigating the failure. It runs ok locally...

dleangen:
It's the dependency check that's off by 1. My local version must be different.

PR #334, opened 2024-01-30
Replace slf4j-simple with logback (v1.4.14)

### LINKED ISSUE #333: Replace slf4j-simple with logbook
Currently, the system uses slf4j-simple as its Logging Provider. This implementation is, as the name suggests, simple. It does not really allow any easy access to configure the logs, at least not programmatically.

By replacing it with a more advanced Logging Provider, sensiNact will have better logging capabilities.

Logback seems to be a reasonable choice.

dleangen:
This PR fixes #333.

There are 2 questions I have before this is merged, though.

First, all the integration tests are running fine. Does this mean that the PR is good to go? Or is there any additional test for the overall working system, if you know what I mean, since we are changing the launcher configuration?

Second, I could not resolve locally without using JavaSE-17, so I updated from JavaSE-11 to JavaSE-17. Is that ok, or do I need to revert back to JavaSE-11 and fix my problem locally?


pom.xml
```diff
@@ -413,10 +419,10 @@
...
       </dependency>
       <dependency>
-        <groupId>org.slf4j</groupId>
-        <artifactId>slf4j-simple</artifactId>
-        <version>${slf4j.version}</version>
-        <scope>runtime</scope>
+        <groupId>ch.qos.logback</groupId>
+        <artifactId>logback-classic</artifactId>
+        <version>${logback.version}</version>
+        <scope>test</scope>
```
timothyjward, line 425: Should this have been left as `runtime` scope?

dleangen:
Hmmmm. The build passes locally. I'm having a bit of trouble deciphering the error message. Maybe related to the JavaSE-17 thing? What do you think, @timothyjward?

dleangen:
> The base version for sensiNact is currently Java 11, and Logback 1.4.x also targets Java 11, so we would prefer to keep the base as JavaSE-11. 

Done.

pom.xml
dleangen: > Should this be runtime scope, and if so shouldn't that be set above on line 327 rather than declaring the dependency twice?

Yes, agree. Now I understand how the `pom.xml` is structured. I made some bad assumptions when I updated the file. Doh! 🤦‍♂️

I can now see that the dependencies are ordered by scope. I moved the dependency to include it only once, along with the other runtime dependencies. 👍

Just waiting for local tests to pass before I commit.

dleangen: Done!

dleangen:
Ok, should be all good to go. When I make changes like that, should I convert the PR to draft mode until the additional work is done?

timothyjward:
> Ok, should be all good to go. When I make changes like that, should I convert the PR to draft mode until the additional work is done?

Yes if you’re not ready for it to be merged

dleangen:
Just added logging configuration for JUnit tests. However, upon running the build again, I realized that the bulk of the logs seems to come from the integration tests, which has to be configured differently.

I will work on that now.

dleangen:
The JUnit tests are using Logback as the Logging Provider, and the configuration is taking effect.

However, it looks like the bulk of the huge quantity of logs is coming from elsewhere, such as Jetty and Aries. I think that these systems are using different logging backends, so the configuration is not taking effect.

I think it is out of scope for me to look into reducing the log output any more than I've already done here, so I'm going to stop here and ask for resumption of merging the PR.

Thanks! 👋

timothyjward (approved):

PR #341, opened 2024-02-01
Better error handling for pushed updates

### LINKED ISSUE #340: Pushed data updates fail silently
The `DataUpdate` service allows asynchronous pushed updates to the Digital Twin, however it isn't good at error handling - particularly with bulk updates, where subsequent updates after a failure are missed.

We need a more robust error handling strategy.

timothyjward:
This commit defines the exceptions that will be used to fail pushed updates, including bulk updates, and will be used to fail the returned promise. These exceptions return the error, and the original DTO so that users can attempt to work out what they did wrong.

Fixes #340


timothyjward:
This PR has API changes in it, so I'd like to get some extra eyes on it please! @tcalmant @juergen-albert @stbischof

tcalmant (approved):

juergen-albert:
+1 from me

PR #348, opened 2024-02-08
Device factory updates

tcalmant:
This PR enhance the usability of the Device Factory:

* Added support for other primitive types: bytes, char, short and boolean
* Parsed arrays (`byte[]`) are now stored as immutable lists
* Enhanced date time parsing options (added support for locale-based long string, documentation was updated)
* Added a `log.errors` flag to the mapping options to ease debugging by tracing parsing, mapping and digital twin update errors
* The HTTP device factory now logs exceptions encountered, as some SSL connection errors were just indicating "status code 0" with no other details.

### DIFF
--- a/southbound/device-factory/device-factory-core/src/main/java/org/eclipse/sensinact/gateway/southbound/device/factory/ValueType.java
+++ b/southbound/device-factory/device-factory-core/src/main/java/org/eclipse/sensinact/gateway/southbound/device/factory/ValueType.java
@@ -33,11 +33,11 @@ public enum ValueType {
     /**
      * Represent available types
      */
-    AS_IS("any", null),
-    STRING("string", (v, o) -> String.valueOf(v)),
+    AS_IS("any", null, null),
+    STRING("string", (v, o) -> String.valueOf(v), String.class),
     CHAR("char", (v, o) -> {
         if (v instanceof Character) {
-            return v;
+            return (Character) v;
         } else if (v instanceof CharSequence) {
             final String str = v.toString();
             if (!str.isEmpty()) {
@@ -55,10 +55,10 @@ public enum ValueType {
         } else {
             return null;
         }
-    }),
+    }, Character.class),
     BOOLEAN("boolean", (v, o) -> {
         if (v instanceof Boolean) {
-            return v;
+            return (Boolean) v;
         } else if (v instanceof Number) {
             return ((Number) v).longValue() != 0;
         } else if (v instanceof CharSequence) {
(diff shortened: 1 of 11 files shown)


southbound/device-factory/device-factory-core/src/main/java/org/eclipse/sensinact/gateway/southbound/device/factory/ValueType.java
```diff
@@ -26,38 +26,69 @@
...
      * Represent available types
      */
-    AS_IS("any", null),
-    STRING("string", (v, o) -> String.valueOf(v)),
-    INT("int", (v, o) -> {
+    AS_IS("any", null, null), STRING("string", (v, o) -> String.valueOf(v), String.class), CHAR("char", (v, o) -> {
+        if (v instanceof Character) {
+            return (Character) v;
+        } else if (v instanceof CharSequence) {
+            final String str = v.toString();
```
timothyjward, line 38: We don't need to use a String here. Cast to CharSequence and then check for length `0`.

southbound/device-factory/device-factory-core/src/main/java/org/eclipse/sensinact/gateway/southbound/device/factory/impl/FactoryParserHandler.java
```diff
@@ -194,15 +202,28 @@ public void handle(final DeviceMappingConfigurationDTO configuration, final Map<
...
                     }
                 }
 
                 if (!bulk.dtos.isEmpty()) {
                     // Send all updates to the gateway thread at once
-                    dataUpdate.pushUpdate(bulk);
+                    Promise<?> pushUpdate = dataUpdate.pushUpdate(bulk);
+                    if (logErrors) {
+                        pushUpdate = pushUpdate.onFailure((t) -> logger
+                                .error("Error updating digital twin with parser {}: {}", parserId, t.getMessage(), t));
```
timothyjward, line 216: I'd suggest doing a walk through of the failed updates if this is `FailedUpdatesException`

southbound/device-factory/device-factory-core/src/main/java/org/eclipse/sensinact/gateway/southbound/device/factory/impl/FactoryParserHandler.java
```diff
@@ -913,15 +1015,14 @@ private Instant parseDateTime(String strDateTime, ZoneId timezone, String format
      */
     private Instant convertTimestamp(final String provider, final Long timestamp) {
 
-        int currentLogMs = (int) Math.log10(System.currentTimeMillis());
-        int currentLogNs = (int) Math.log10(System.nanoTime());
+        int log10ms = (int) Math.log10(System.currentTimeMillis());
         int timestampLog = (int) Math.log10(timestamp);
 
-        if (timestampLog == currentLogMs) {
+        if (timestampLog == log10ms) {
```
timothyjward, line 1021: I think this will have issues around base 10 rollover. Would:

    if (timestampLog > (log10ms + 2)) {
        // Over 500 years in the future - guess nanos
        return Instant.EPOCH.plusNanos(timestamp);
    } else if (timestampLog < (log10ms - 1) {
        // Over 45 years in the past, guess seconds
        return Instant.ofEpochSecond(timestamp);
    } else {
        return Instant.ofEpochMilli(timestamp);
    }

be better?

southbound/device-factory/device-factory-core/src/test/java/org/eclipse/sensinact/gateway/southbound/device/factory/impl/RecordHandlingTest.java
```diff
@@ -310,4 +334,94 @@ void testVariableService() throws Exception {
...
+            bulks.clear();
+            config.mapping.put("@datetime", "locale_" + locale.getLanguage());
+            config.mappingOptions.formatDateTime = null;
+            config.mappingOptions.formatDateTimeLocale = locale.toString();
+            config.mappingOptions.formatDateStyle = "long";
+            config.mappingOptions.formatTimeStyle = "medium";
+            config.mappingOptions.dateTimezone = "+01";
+            deviceMapper.handle(config, Map.of(), new byte[0]);
+            dto = getResourceValue("provider", "data", "val");
+            assertEquals(now.truncatedTo(ChronoUnit.SECONDS), dto.timestamp);
```
timothyjward, line 424: I think that this is vulnerable to rollover failures.
```suggestion
            long delta = Math.abs(ChronoUnit.SECONDS.between(Instant.now(), dto.timestamp));
            assertTrue(delta < 2, "The dto timestamp " + dto.timestamp + " is skewed from " + Instant.now());
```

timothyjward (approved):

southbound/device-factory/device-factory-core/src/test/java/org/eclipse/sensinact/gateway/southbound/device/factory/impl/RecordHandlingTest.java
tcalmant: Not sure on that one: we're comparing the same instant

southbound/device-factory/device-factory-core/src/main/java/org/eclipse/sensinact/gateway/southbound/device/factory/impl/FactoryParserHandler.java
tcalmant: Is it something we should handle or should we suppose the JVM will manage it?
` Instant.ofEpochSecond(Instant.MAX.getEpochSecond())` works, with Instant.MAX being end of year 1_000_000_000.

timothyjward: I don't mean rollover of the epoch, just the seconds representation. If Instant.now() is `1000_000_000` seconds and the update time is `999_999_999` seconds then the `log10` won't match the `timestampLog`. I think we need slightly fuzzy matching (e.g. allowing +/-1) to allow for this scenario.

southbound/device-factory/device-factory-core/src/test/java/org/eclipse/sensinact/gateway/southbound/device/factory/impl/RecordHandlingTest.java
timothyjward: Ok - I wasn't sure if we had some auto-generated timestamps. If it's actually the same time being used then the existing check is great.

southbound/device-factory/device-factory-core/src/main/java/org/eclipse/sensinact/gateway/southbound/device/factory/impl/FactoryParserHandler.java
tcalmant: OK I see, I'll apply the change then

PR #361, opened 2024-03-07
Add back gogo commands

dleangen:


dleangen:
@timothyjward and @tcalmant, if you have the time, please take a look at this first iteration.

This is a work in progress. Just looking for some initial feedback. I hope to wrap this up tomorrow (today for you).

Thanks!

tcalmant:
Hi @dleangen,

I think the overall idea is correct, here are my remarks:
* I'm not fond of the string cast (`return (T)"<NULL>";`): wouldn't it be better to return null?
* I don't think the use of the JSON API is necessary, but if you need to parse JSON, you should use the Jackson ObjectMapper instead: it's already a dependency as it's used in many places in sensiNact (see [here](https://github.com/eclipse/org.eclipse.sensinact.gateway/blob/969d78788b379e093a51fea516e8933c8710674e/core/impl/src/main/java/org/eclipse/sensinact/core/model/nexus/emf/EMFUtil.java#L79) for a usage example). 
* You might want to install the EditorConfig plugin for your IDE then use its formatting options to avoid format check errors
* Be careful of the copyright year in new files

It's been a while since I last wrote a Gogo shell command, I'm not sure how interactive it can be:
* subscribe is a good idea if it can be stopped easily (blocking command displaying updated until a key is pressed? command returning the subscription ID to give to "unsubscribe"?)
* the user login will require to enter a password/secret, if we want to support it

dleangen:
Thanks for your comments, @tcalmant!

> I'm not fond of the string cast (return (T)"`<NULL>`";): wouldn't it be better to return null?

When no object is found, I wanted to return some kind of message to the user. The simplest way, since gogo is pretty rudimentary, is to just return a `String`. I agree that the message could be better than "`<NULL>`". In essence, the return value is something like "`T or String`" (`T` if there is a "good" answer, or `String` if there is not a "good" answer). To make the return value consistent, if that is the biggest concern, I could return `String` in all cases and cast the return object using `toString()`, but other than having a consistent return value, I think it ends up being the same old thing. I think it's just the nature of how the gogo commands work, which is why I'm not so worried about it. Gogo doesn't care what you give it: it all gets converted to `String` anyway.  Now, I could throw an exception, but the message returned on the command line is not very friendly. I don't think semantically it's the right thing to do anyway.

So, the way I did it allows me to control the message that the user sees, with the disadvantage that there is the strange cast that you see.

> I don't think the use of the JSON API is necessary

You're right, it's not. It will be removed. It was just there as a consequence of the legacy version of the commands, which used JSON. Thanks for pointing it out, though.

> You might want to install the EditorConfig plugin for your IDE then use its formatting options to avoid format check errors.

Righto! I actually tried that before, but it ended up being complicated and I gave up. I'll look into it again.

> Be careful of the copyright year in new files

I suppose that the plugin will catch those?

> subscribe is a good idea if it can be stopped easily (blocking command displaying updated until a key is pressed? command returning the subscription ID to give to "unsubscribe"?) the user login will require to enter a password/secret, if we want to support it

I indeed plan on including authentication. I haven't yet looked at subscribe. I will take these comments into consideration when I do.

stbischof:
May add

@org.osgi.annotation.bundle.Capability( 
     namespace = "org.apache.felix.gogo", 
     name = "command.implementation", 
     version = "1.0.0" 
 ) 
 @org.osgi.annotation.bundle.Requirement( 
     effective = "active", 
     namespace = "org.apache.felix.gogo", 
     name = "runtime.implementation", 
     version = "1.0.0" 
 )

stbischof:
I would Suggestion to Return more dto types then formatted Strings.

A formatter can di this job better

https://github.com/aQute-os/biz.aQute.osgi.util/tree/master/biz.aQute.gogo.commands.provider%2Fsrc%2Fmain%2Fjava%2Fbiz%2FaQute%2Fgogo%2Fcommands%2Fdtoformatter

dleangen:
@stbischof Thank you for your comments.

> May add
> @org.osgi.annotation.bundle.Capability
> @org.osgi.annotation.bundle.Requirement

Excellent suggestion! I have added those.

> I would Suggestion to Return more dto types then formatted Strings.
> A formatter can di this job better

I agree in principle, but the docs are sparse. Do you know of any examples I could emulate?

Else, I would propose to use the less-than-ideal `String` for now, and have the formatting as a separate issue.

stbischof:
You may have used the scr commands

https://github.com/apache/felix-dev/blob/master/scr%2Fsrc%2Fmain%2Fjava%2Forg%2Fapache%2Ffelix%2Fscr%2Fimpl%2FComponentCommands.java#L143-L149

stbischof (commented):
rm space

dleangen:
> rm space

Which space?

If you are referring to the spaces in the `@Descriptor` text, that is because I wanted to align the text in the help output. If you run `help`, you'll notice that the text lines up very nicely.

stbischof (commented):
returns needes ? Espacially at the end?

stbischof:
Sorry doing this with GitHub app. This should be comments in code.

dleangen:
> Sorry doing this with GitHub app. This should be comments in code.

lol. Ok, that now makes sense. Thank you very much for taking the time to comment on the code. I don't yet see any comments in the code. I suppose I need to wait for that to appear?

stbischof:
I dont think that the code links will reappear.

But it was just about space and /n usage.

dleangen:
@stbischof I am having trouble parsing some of your comments above. 🤣

I am, however, looking into the `Converter`s. That's pretty cool! I had been using `Converter`s to format input parameters, but never before for the output. I always wondered what that method was for. Now I know! I agree that it cleans up the code quite a lot, and is worth doing.

I have a question: in some cases, the reply to the query is `null`. When that happens, when using a `Converter`, there is no output. I would prefer to output something like "No Provider Found".

How can I handle cases like this using a `Converter`?

stbischof:
Using Optional would be the right was for an api to declare that not always you will get something back. This way get rid of the null Problem and handle the optional.

dleangen:
> Using Optional would be the right was for an api to declare that not always you will get something back. This way get rid of the null Problem and handle the optional.

@stbischof Thanks! That would have to be done in the command method itself, which is doable. However, I believe, unless I am misunderstanding a nuance from a previous similar discussion, that @timothyjward would not agree.

stbischof:
@timothyjward mostly has very good Arguments.
Interested in hearing this.

platform/sensinact-shell/src/main/java/org/eclipse/sensinact/gateway/commands/gogo/SensiNactCommandSession.java
```diff
@@ -0,0 +1,61 @@
...
+    @Activate
+    void activate() {
+        session = sessionManager.createNewAnonymousSession();
+    }
+
+    /**
+     * Returns the command session. Value will never be null;
+     */
+    public SensiNactSession get() {
+        // TODO: Can the session expire?
```
timothyjward, line 49: Yes, sessions can and do expire

https://github.com/eclipse/org.eclipse.sensinact.gateway/blob/969d78788b379e093a51fea516e8933c8710674e/core/api/src/main/java/org/eclipse/sensinact/core/session/SensiNactSession.java#L39-L61

pom.xml
```diff
@@ -97,6 +97,7 @@
 
   <modules>
     <module>core</module>
+    <module>platform</module>
```
timothyjward, line 100: What was the reasoning behind creating a new top-level module? Aren't these commands behaving as another type of Northbound provider?

timothyjward: > I agree that for this first release, it only interacts with northbound providers.

I would go a little further and say that it **is** a northbound provider, in that it is using the mechanisms other northbound providers do to provide access to data from the sensiNact core.

> Do you see this as eventually expanding to interface with southbound providers as well?

I think that if we create commands which are doing the job of a southbound provider I would expect them to be compiled and packaged separately, and that those commands would live in the `southbound` module.

If we start having commands that do other things (not that I can think what those might be at this time) then at that point I can see that we might start a separate top level module for the build.

dleangen: Ok. In that case, I'll remove the `platform` package and find a different home. Any preferences?

timothyjward: Definitely in the `northbound` folder

platform/sensinact-shell/src/main/java/org/eclipse/sensinact/gateway/commands/gogo/SensiNactCommandSession.java
dleangen: Thanks! Any advice as to how long to extend for? I was thinking an hour...

dleangen:
I believe that I have addressed all concerns, and that this PR is ready to merge. 🙏

timothyjward (approved):

PR #369, opened 2024-03-28
Move the Session API into the Northbound project

timothyjward:
User Sessions are really part of the northbound access mechanism, and should not be part of the core. This will simplify adding access restrictions in the future.


timothyjward:
This commit involves a change of package name, but I'm convinced that it's a necessary structural improvement which will allow the introduction of access checks in the session without impacting the core, which needs to stay simple and fast.

tcalmant (approved):

PR #390, opened 2024-04-29
Fix/notification service map

gg-dim:
Service updates for services in dynamic providers were ignored.


stbischof:
Could you reduce the 3 Format commits ?  Seems that only 4 lines changed

PR #392, opened 2024-04-30
Fix change notification for services in dynamic provider

gg-dim:
Changes from services in dynamic providers (services map) were ignored.


timothyjward:
Is there any chance of a test case for this change? It looks useful, but I worry about regressions if we don't have a test which covers the behaviour.

juergen-albert:
The testcase will come with the next PR, we already have in preparation

PR #393, opened 2024-05-06
Dynamic service

juergen-albert:


juergen-albert:
@gg-dim Can you fix the licence and Editor Config Errors.

PR #397, opened 2024-05-20
Include the model name in topic notifications

timothyjward:
Filtering at the model level is extremely useful as it permits efficient tracking of multiple providers using the same model. The model URI is required to ensure strict uniqueness, however the model uri is difficult to represent in topic syntax, and would need large amounts of escaping. The trade off is therefore to allow the model name to be used in the topic for rapid checking, with the filter being available for more precise matching if needed.


tcalmant (approved):
LGTM

PR #410, opened 2024-06-18
Reinstate final modifier for GenericDto

timothyjward:
The EMF dto altered the GenericDto so that it could extend it. The GenericDto was intentionally final to prevent usage mistakes in the API. There are few fields and we should copy them.


juergen-albert:
Okay, makes sense. What you did, seems to be fine as well.

PR #415, opened 2024-06-25
Added metrics provider

tcalmant:
* If the `provider.enabled` option of `sensinact.metrics` is true, the metrics bulk DTO is pushed to the gateway.
* Added a test to ensure the provider exists
* Fixed the JVM gauge sometimes not being registered


tcalmant:
Looks like the whiteboard metrics name must be normalized

timothyjward (approved):

PR #420, opened 2024-07-08
Extend whiteboard capabilities

tcalmant:
This PR adds support for new resources handling whiteboard services that doesn't rely on method annotations.
This eases the creation of resources and handlers on-the-go.

The existing whiteboard API with annotated methods hasn't been modified.

### DIFF
--- a/core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/AbstractDescriptiveReadOnly.java
+++ b/core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/AbstractDescriptiveReadOnly.java
@@ -1,0 +1,48 @@
+/*********************************************************************
+* Copyright (c) 2024 Kentyou.
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which is available at https://www.eclipse.org/legal/epl-2.0/
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*   Thomas Calmant (Kentyou) - initial implementation
+**********************************************************************/
+package org.eclipse.sensinact.core.whiteboard;
+
+import java.lang.reflect.ParameterizedType;
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+
+import org.eclipse.sensinact.core.twin.TimedValue;
+import org.osgi.util.promise.Promise;
+import org.osgi.util.promise.PromiseFactory;
+
+public abstract class AbstractDescriptiveReadOnly<T> implements WhiteboardGet<T>, WhiteboardResourceDescription<T> {
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public Class<T> getResourceType() {
+        return (Class<T>) ((ParameterizedType) getClass().getGenericSuperclass()).getActualTypeArguments()[0];
+    }
+
+    @Override
+    public Duration getCacheDuration() {
+        return Duration.of(30, ChronoUnit.SECONDS);
+    }
+
+    @Override
+    public Promise<TimedValue<T>> pullValue(PromiseFactory pf, String modelPackageUri, String model, String provider,
+            String service, String resource, Class<T> resourceType, TimedValue<T> cachedValue) {
+        Promise<TimedValue<T>> p = doPullValue(pf, modelPackageUri, model, provider, service, resource, cachedValue);
+        if (p == null) {
+            return pf.failed(new NullPointerException(getClass().getName() + " returned a null promise"));
+        }
+        return p;
+    }
+
+    protected abstract Promise<TimedValue<T>> doPullValue(PromiseFactory promiseFactory, String modelPackageUri,
+            String model, String provider, String service, String resource, TimedValue<T> cachedValue);
+}
--- a/core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/WhiteboardConstants.java
+++ b/core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/WhiteboardConstants.java
@@ -1,0 +1,73 @@
+/*********************************************************************
+* Copyright (c) 2024 Kentyou.
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which is available at https://www.eclipse.org/legal/epl-2.0/
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*   Thomas Calmant (Kentyou) - initial implementation
+**********************************************************************/
+package org.eclipse.sensinact.core.whiteboard;
+
+public interface WhiteboardConstants {
+
+    /**
+     * Property held by whiteboard handler service to indicate the list of providers
+     * it supports to be called for.
+     *
+     * Property can be null, a string or a collection or array of strings.
+     *
+     * If the property is null or an empty collection, the handler accepts all
+     * providers.
+     *
+     * If multiple handlers match the provider, the one with the smallest non-empty
+     * list of providers is preferred
+     */
+    String PROP_PROVIDERS = "sensiNact.provider.name";
+
+    /**
+     * Property held by whiteboard handler service to indicate the model package URI
+     * it supports to be called for.
+     *
+     * Property is a string that can be null, in which case it is computed based on
+     * default package and the given model name
+     */
+    String PROP_MODEL_PACKAGE_URI = "sensiNact.whiteboard.modelPackageUri";
+
+    /**
+     * Property held by whiteboard handler service to indicate the name of the model
+     * it supports to be called for.
+     *
+     * Property is a string that can't be null
+     */
+    String PROP_MODEL = "sensiNact.whiteboard.model";
+
+    /**
+     * Property held by whiteboard handler service to indicate the name of the
+     * service it supports to be called for.
+     *
+     * Property can be null to indicate any service
+     */
+    String PROP_SERVICE = "sensiNact.whiteboard.service";
+
+    /**
+     * Property held by whiteboard handler service to indicate the name of the
+     * resource it supports to be called for.
+     *
+     * Property is a string that can be null to indicate any resource
+     */
+    String PROP_RESOURCE = "sensiNact.whiteboard.resource";
+
+    /**
+     * Property held by whiteboard handler service to indicate that it requires the
+     * whiteboard to create the resource. Value handlers (GET/SET) must implement
+     * {@link WhiteboardResourceDescription} and action handlers must implement
+     * {@link WhiteboardActDescription} to be able to describe their resource.
+     *
+     * Property is a boolean that can be null.
+     */
+    String PROP_AUTO_CREATE = "sensiNact.whiteboard.create";
+}
--- a/core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/WhiteboardConstants.java
+++ b/core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/WhiteboardConstants.java
@@ -49,7 +49,8 @@ public interface WhiteboardConstants {
      * Property held by whiteboard handler service to indicate the name of the
      * service it supports to be called for.
      *
-     * Property can be null to indicate any service
+     * Property can be null to indicate any service if and only if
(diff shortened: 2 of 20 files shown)


core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/WhiteboardGetSet.java
```diff
@@ -0,0 +1,68 @@
...
+import org.eclipse.sensinact.core.twin.TimedValue;
+import org.osgi.util.promise.Promise;
+import org.osgi.util.promise.PromiseFactory;
+
+/**
+ * Whiteboard service to handle resource get/set operations
+ *
+ * @param <T> Resource type
+ */
+public interface WhiteboardGetSet<T> {
```
timothyjward, line 26: I think that these should probably be separate interfaces. My reasoning is that there are a lot of read-only resources and the only way I can see to indicate this is by throwing `UnsupportedOperationException`.

It would be possible to make `WhiteboardSet` extend `WhiteboardGet` and for the default `pullValue` method of `AbstractWhiteboardSet` to just return the cached value.

core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/WhiteboardGetSet.java
```diff
@@ -0,0 +1,68 @@
...
+     * @param promiseFactory  Promise factory
+     * @param modelPackageUri Package URI of the provider model
+     * @param model           Provider model name
+     * @param provider        Provider name
+     * @param service         Service name
+     * @param resource        Resource name
+     * @param cachedValue     Previously cached value
+     * @return A promise, created with the promise factory
+     */
+    Promise<TimedValue<T>> pullValue(PromiseFactory promiseFactory, String modelPackageUri, String model,
```
timothyjward, line 50: The annotated version allows you to request the target type as a parameter, which I think that we should include here.

core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/WhiteboardGetSet.java
```diff
@@ -0,0 +1,68 @@
...
+     * @param modelPackageUri Package URI of the provider model
+     * @param model           Provider model name
+     * @param provider        Provider name
+     * @param service         Service name
+     * @param resource        Resource name
+     * @param cachedValue     Previously cached value
+     * @param newValue        New value
+     * @return A promise, created with the promise factory
+     */
+    Promise<TimedValue<T>> pushValue(PromiseFactory promiseFactory, String modelPackageUri, String model,
```
timothyjward, line 66: The annotated version allows you to request the target type as a parameter, which I think that we should include here.

core/impl/src/main/java/org/eclipse/sensinact/core/whiteboard/impl/SensinactWhiteboard.java
```diff
@@ -133,6 +135,48 @@ public SensinactWhiteboard(GatewayThread gatewayThread, IMetricsManager metrics)
         this.metrics = metrics;
     }
 
+    private Map<RegistryKey, WhiteboardContext<WhiteboardGetSet<?>>> rcHandlers = new HashMap<>();
```
timothyjward, line 138: Would it make more sense to introduce these as `ActMethod`, `GetMethod` and `SetMethod` types so that we only have one registry to check each time we do anything?

core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/WhiteboardGetSet.java
tcalmant: I'm not sure it's necessary here, as the type comes from the class implementing pullValue itself to define the resource.
In the annotated case, it's useful as the type might not be easily accessible otherwhise.

core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/WhiteboardGetSet.java
tcalmant: I'm not sure it's necessary here, as the type comes from the class implementing pullValue itself to define the resource.
In the annotated case, it's useful as the type might not be easily accessible otherwise.

core/impl/src/main/java/org/eclipse/sensinact/core/command/impl/GatewayThreadImpl.java
```diff
@@ -128,6 +130,24 @@ void removeEPackage(EPackage ePackage) {
...
+
+    void removeWhiteboardResourceHandler(WhiteboardGetSet<?> getHandler, Map<String, Object> props) {
+        whiteboard.removeWhiteboardHandler(getHandler, props);
+    }
+
+    @Reference(service = WhiteboardAct.class, cardinality = MULTIPLE, policy = DYNAMIC)
+    void addWhiteboardActHandler(WhiteboardAct<?> actHandler, Map<String, Object> props) {
+        whiteboard.addWhiteboardActHandler(actHandler, props);
+    }
+
```
timothyjward, line 146: Was there a reason to avoid managing service update?

core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/WhiteboardGetSet.java
tcalmant: That was my initial design, but I preferred to define an Abstract with a read-only behaviour by default as it also provides the detection of the resource type by reflection.
I'm not yet sure what's the easiest API here

core/impl/src/main/java/org/eclipse/sensinact/core/command/impl/GatewayThreadImpl.java
tcalmant: Still a work in progress

core/impl/src/main/java/org/eclipse/sensinact/core/whiteboard/impl/SensinactWhiteboard.java
timothyjward: On reflection I think that there might be some benefit in extracting interfaces for `GetMethod`, `SetMethod` and `ActMethod` which we use internally to abstract between annotated and non-annotated whiteboard services, the existing implementations then work for annotated methods and new (simple) implementations work for whiteboard services.

We could then use the same updated method for all types in the SensinactWhitebaord

core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/WhiteboardGetSet.java
timothyjward: I think separating them is important, otherwise the runtime has no way to know if a resource should be modelled as modifiable.

tcalmant: Also, I'm not sure if the resource type / action result type should be from the handler itself or from its services properties.
Having it in the properties could allow defining handlers using lambda expressions

core/impl/src/main/java/org/eclipse/sensinact/core/whiteboard/impl/SensinactWhiteboard.java
tcalmant: That was the initial plan, but returned types are not exactly the same (e.g. value instead of promise of value for actions)

timothyjward: The return is deliberately fuzzy for annotated, it can be a promise or not: 

https://github.com/eclipse/org.eclipse.sensinact.gateway/blob/fefa2b29b52e2339061cd82f296a173818365d7d/core/impl/src/main/java/org/eclipse/sensinact/core/whiteboard/impl/SensinactWhiteboard.java#L506-L510 

The interface versions will just be less fuzzy

core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/WhiteboardHandler.java
```diff
@@ -0,0 +1,20 @@
...
+*
+* Contributors:
+*   Thomas Calmant (Kentyou) - initial implementation
+**********************************************************************/
+package org.eclipse.sensinact.core.whiteboard;
+
+/**
+ * Common parent interface for whiteboard handlers
+ */
+public interface WhiteboardHandler<T> {
```
timothyjward, line 18: I don't think that this needs generics - it also makes things difficult if you implement get/set/act in a single service as you can have different type hierarchies for the same interface with competing type parameters.

core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/AbstractDescriptiveReadOnly.java
```diff
@@ -0,0 +1,48 @@
...
+import org.eclipse.sensinact.core.twin.TimedValue;
+import org.osgi.util.promise.Promise;
+import org.osgi.util.promise.PromiseFactory;
+
+public abstract class AbstractDescriptiveReadOnly<T> implements WhiteboardGet<T>, WhiteboardResourceDescription<T> {
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public Class<T> getResourceType() {
+        return (Class<T>) ((ParameterizedType) getClass().getGenericSuperclass()).getActualTypeArguments()[0];
```
timothyjward, line 28: This is fine for now, but may need to be enhanced to cope with types that don't directly extend this class

core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/WhiteboardConstants.java
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * Property is a string that can't be null
+     */
+    String PROP_MODEL = "sensiNact.whiteboard.model";
+
+    /**
+     * Property held by whiteboard handler service to indicate the name of the
+     * service it supports to be called for.
+     *
+     * Property can be null to indicate any service
```
timothyjward, line 52: I think that this is only possible if the `PROP_RESOURCE` is also `null`?

core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/WhiteboardResourceDescription.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.time.Duration;
+
+/**
+ * Interface that an auto-descriptive value resource handler must implement if
+ * it expects the whiteboard to create its resource automatically
+ *
+ * @param <T> Type of resource
+ */
+public interface WhiteboardResourceDescription<T> {
```
timothyjward, line 23: Using 
```suggestion
public interface WhiteboardResourceDescription<T> extends WhiteboardGet<T> {
```
Would force the reified type of the the description and the get to be the same, or we could use an indirection to force the type variable of the description to be a super type of the type variable for the get.

core/impl/src/test/java/org/eclipse/sensinact/core/command/impl/WhiteboardImplTest.java
```diff
@@ -738,4 +754,542 @@ void testPushPull() throws Throwable {
...
+            thread.addWhiteboardResourceHandler(getHandler, makeProps(modelName, svcName, "rc"));
+
+            createProviders(modelName, svcName);
+
+            TimedValue<Integer> result = getValue(PROVIDER_A, svcName, "rc", Integer.class);
+            assertNotNull(result.getValue(), "No value");
+            assertNotNull(result.getTimestamp(), "No timestamp");
+            assertEquals(42, result.getValue());
+
+            // FIXME: Change exception type
```
timothyjward, line 790: Fix this?

timothyjward (approved):
Minor comments, but otherwise looks really good

core/api/src/main/java/org/eclipse/sensinact/core/whiteboard/AbstractDescriptiveReadOnly.java
tcalmant: Agreed

PR #426, opened 2024-07-23
Fix to list services and resources on newly created provider instances

tcalmant:
... else services when no value was explicitly set are not snapshotted


timothyjward:
We found an issue when trying to snapshot newly created providers, in that the snapshots had no values for some of the services (rather than resources which aren't set).

There were two ways to solve this. One was to eagerly create the services as soon as the provider was created, the other was to be more careful about handling null service instances.

This fix attempts to do the latter. It is more invasive to the core than eagerly creating the services, but the changes do not go beyond the module boundary. The former would have affected the event delivery and models, and would therefore have been visible externally.

timothyjward:
@juergen-albert - it would be good to get your opinion of this as it touches the EMF model in a few places.

timothyjward:
The dash error was fixed in #428

PR #427, opened 2024-07-23
Allow setting metadata to a resource with no value set yet

tcalmant:


timothyjward (approved):

timothyjward:
It turns out that @juergen-albert is away this week, and there are a few issues in this area. We're going to go ahead and merge, scheduling a review of the overall core changes with the various bug fixes in place.

PR #425, opened 2024-07-23
Snapshot API enhancement

tcalmant:
Added utility methods to ease handling of snapshots

PR #430, opened 2024-07-25
Implement the expand query for SensorThings northbound

### LINKED ISSUE #418: Feature Request Sensor Thing API Extending Responses with nested Entities
would be nice to implement $expand :
- https://docs.ogc.org/is/18-088/18-088.html#select3

timothyjward:
This commit adds a backing implementation for the expand query for the SensorThings REST API. This is a pervasive change which requires us to carry context throughout the query so as to attach additional data structures based on the filter content.

* All queries now use full ProviderSnapshot instances so that we don't have to re-query for other resource data when filling in expansions
* Limited bulk data is included in expansions to prevent the JSON size exploding
* Data context is mapped by id to ensure we can expand values in lists
* All context is carried outside the DTO to avoid unnecessary fields appearing in the JSON output

Note that we require that all ids are able to be turned into strings. This is currently true throughout the gateway.

Fixes #418


northbound/sensorthings/rest.gateway/src/main/java/org/eclipse/sensinact/sensorthings/sensing/rest/impl/AbstractAccess.java
```diff
@@ -0,0 +1,109 @@
...
+        return es == null ? EMPTY : es;
+    }
+
+    private Optional<ProviderSnapshot> getProviderSnapshot(String id) {
+        return getSession().filteredSnapshot(new SnapshotFilter(id)).stream().findFirst();
+    }
+
+    protected Optional<ResourceSnapshot> getResourceSnapshot(ProviderSnapshot providerSnapshot,
+            String service, String resource) {
+        return providerSnapshot.getServices().stream()
```
tcalmant, line 76: Since #425, there is a providerSnapshot.getResource(service, resource) method

tcalmant (approved):
LGTM
AbstractAccess.getResourceSnapshot could be simplified to call ProviderSnapshot.getResource

timothyjward:
> LGTM AbstractAccess.getResourceSnapshot could be simplified to call ProviderSnapshot.getResource

I removed it completely and inlined the one place it was used with `ProviderSnapshot.getResource`

PR #431, opened 2024-07-25
Fixed inclusion of metadata in resource description

tcalmant:
Metadata were never added to the resource description.
Exclude timestamp as it is associated to resource value (GET)

### DIFF
--- a/northbound/query-handler/src/main/java/org/eclipse/sensinact/northbound/query/impl/QueryHandler.java
+++ b/northbound/query-handler/src/main/java/org/eclipse/sensinact/northbound/query/impl/QueryHandler.java
@@ -591,31 +591,28 @@ public class QueryHandler implements IQueryHandler {
     }
 
     /**
-     * Generates the list of resource metadata
+     * Generates the list of resource metadata, excluding timestamp (shown alongside
+     * the resource value)
      *
      * @param resource Resource to describe
      * @param metadata Raw resource metadata
-     * @return Description of resource metadta
+     * @return Description of resource metadata
      */
-    private List<MetadataDTO> generateMetadataDescriptions(final SensinactResource resource,
-            Map<String, Object> metadataMap) {
-        final List<MetadataDTO> result;
+    private List<MetadataDTO> generateMetadataDescriptions(Map<String, Object> metadataMap) {
         if (metadataMap != null) {
-            result = new ArrayList<>(metadataMap.size());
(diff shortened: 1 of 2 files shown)


northbound/query-handler/src/main/java/org/eclipse/sensinact/northbound/query/impl/QueryHandler.java
```diff
@@ -792,15 +789,15 @@ protected Promise<ResponseDescribeResourceDTO> call(SensinactResource resource,
                         dto.type = resource.getResourceType();
                         dto.accessMethods = generateAccessMethodsDescriptions(resource);
                         return resource.getMetadataValues().then(metadata -> {
-                            dto.attributes = generateMetadataDescriptions(resource, metadata.getValue());
+                            dto.attributes = generateMetadataDescriptions(metadata.getValue());
                             return pf.resolved(dto);
                         });
                     }
                 }).then((d) -> {
```
timothyjward, line 796: Probably not for this PR, but we should be using `map` here as there's no async needed. We should be using `recover` at the end as well.

northbound/query-handler/src/test/java/org/eclipse/sensinact/northbound/query/test/integration/DescriptionsTest.java
```diff
@@ -252,6 +254,29 @@ void resourceDescription() throws Exception {
...
+        dto.type = null;
+        dto.value = null;
+        push.pushUpdate(dto).getValue();
+
+        // Check the resource
+        rawResult = handler.handleQuery(session, query);
+        utils.assertResultSuccess(rawResult, EResultType.DESCRIBE_RESOURCE, PROVIDER, SERVICE, RESOURCE);
+
+        result = (TypedResponse<?>) rawResult;
+        utils.assertResultSuccess(result, EResultType.DESCRIBE_RESOURCE, PROVIDER, SERVICE, RESOURCE);
```
timothyjward, line 272: Is this the same check as before?

timothyjward (approved):

northbound/query-handler/src/main/java/org/eclipse/sensinact/northbound/query/impl/QueryHandler.java
tcalmant: At some point we might need an overall review on Promises handling

PR #445, opened 2024-08-08
dataUpdate now supports multiple updates

gathanase:


timothyjward (approved):
This is a very clean solution. 👍

PR #504, opened 2024-11-26
Using a ecliptic curve private key for the MQTT 

gg-dim:
Using a elliptic curve  private key (auth.clientcert.key.algorithm=EC) for the MQTT client results in an base64 encoding error as the RegEx for cutting the header (----BEGIN PRIVATE KEY-----) and footer (-----END PRIVATE KEY-----) didn't work.


gg-dim:
@timothyjward Something strange happened. Can you please take a look at the package declaration of the new test case. How can a package ```package src.test.java.org.eclipse.sensinact.gateway.southbound.mqtt.test;``` happen and get through the build process?

PR #523, opened 2024-12-18
Snapshot

gathanase:
Implement the snapshot rest service


northbound/query-handler/src/main/java/org/eclipse/sensinact/northbound/query/impl/QueryHandler.java
```diff
@@ -362,14 +376,81 @@ private AbstractResultDTO handleAct(final SensiNactSession userSession, final Qu
...
+        final ResponseSnapshotDTO result = new ResponseSnapshotDTO();
+
+        result.providers = new HashMap<>();
+        result.uri = "/";
+        result.statusCode = 200;
+
+        for (var filter: query.filter) {
+            ICriterion criterion = resourceSelectorFilterFactory.parseResourceSelector(filter);
+            for (var providerSnapshot: executeFilter(criterion)) {
+                if (criterion.getProviderFilter() == null || criterion.getProviderFilter().test(providerSnapshot)) {
```
timothyjward, line 416: Why are we doing these tests? The results from `executeFilter` should already be filtered correctly.

northbound/rest/src/main/java/org/eclipse/sensinact/northbound/rest/impl/RestNorthbound.java
```diff
@@ -193,6 +196,15 @@ public AbstractResultDTO resourceGet(final String providerId, final String servi
         return handleQuery(query);
     }
 
+    @Override
+    public AbstractResultDTO getSnapshot(final boolean includeMetadata, List<ResourceSelector> filter) {
+        final QuerySnapshotDTO query = new QuerySnapshotDTO();
+        query.uri = new SensinactPath();
+        query.filter = filter;
+        query.includeMetadata = Optional.ofNullable(includeMetadata).orElse(false);
```
timothyjward, line 204: A primitive boolean can never be `null`

```suggestion
        query.includeMetadata = includeMetadata;
```

timothyjward (changes requested):
Some minor tidy up needed, but this is a great addition to the API

northbound/query-handler/src/main/java/org/eclipse/sensinact/northbound/query/impl/QueryHandler.java
```diff
@@ -413,34 +413,32 @@ private AbstractResultDTO handleSnapshot(final SensiNactSession userSession, fin
...
-                                    resourceDTO.name = resourceSnapshot.getName();
-                                    resourceDTO.type = resourceSnapshot.getType().getName();
-                                    resourceDTO.timestamp = resourceSnapshot.getValue().getTimestamp().toEpochMilli();
-                                    resourceDTO.value = resourceSnapshot.getValue().getValue();
-                                    if (query.includeMetadata) {
-                                        resourceDTO.attributes = generateMetadataDescriptions(resourceSnapshot.getMetadata());
-                                    }
-                                    serviceDTO.resources.put(resourceSnapshot.getName(), resourceDTO);
+                for (var serviceSnapshot: providerSnapshot.getServices()) {
+                    if (criterion.getServiceFilter() == null || criterion.getServiceFilter().test(serviceSnapshot)) {
```
timothyjward, line 417: As with the provider level, it really isn't necessary to re-test the snapshots with the criterion filters. This is all done when generating the snapshot.

northbound/query-handler/src/main/java/org/eclipse/sensinact/northbound/query/impl/QueryHandler.java
```diff
@@ -413,34 +413,32 @@ private AbstractResultDTO handleSnapshot(final SensiNactSession userSession, fin
...
+                        for (var resourceSnapshot: serviceSnapshot.getResources()) {
+                            if ((criterion.getResourceFilter() == null || criterion.getResourceFilter().test(resourceSnapshot)) && resourceSnapshot.getValue() != null) {
```
timothyjward, line 419: These tests can also be removed, possibly excluding the `resourceSnapshot.getValue() != null` if we shouldn't return resource values that have never been set.

northbound/query-handler/src/main/java/org/eclipse/sensinact/northbound/query/dto/result/SnapshotResourceDTO.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Resource ID
+     */
+    public String name;
+
+    /**
+     * Value milliseconds timestamp
+     */
+    public long timestamp;
```
timothyjward, line 30: Would it be more sensible to leave this as an `Instant`
```suggestion
    public Instant timestamp;
```

northbound/query-handler/src/main/java/org/eclipse/sensinact/northbound/query/impl/QueryHandler.java
gathanase: I thought the criterion was used only to select the provider. I will fix

gathanase:
All requested changes have been applied

timothyjward (approved):

Issue #515, opened 2024-12-10
Support Java Records


timothyjward:
It feels like time to move to Java 17 as a base. Java 11 is EoL, and there are some fun things that we can do with record types:

1. Support Records as pushed updates - this will involve testing/updating the existing annotations for mapping
2. Use records for notifications - records are immutable which is a much better fit for our notifications and will avoid the need for copying.
    * This will involve record support in our Event Bus implementation, which currently exists in the snapshots

Issue #496, opened 2024-11-14
API tidy up


timothyjward:
We are cleaning down before attempting a milestone release. This means that we don't need a final release API, but it would be good to make sure things are in a good shape first.

This issue is to gather changes that may need to be made.

timothyjward:
The `NotificationAccumulator` interface is available from the `GatewayThread` and present in the `AbstractSensinactCommand`. This is an important type but I can't understand the need for it to be public API. Would we ever expect a user to call any of the methods? If they did it could cause chaos by sending events with unvalidated changes....

@tcalmant @juergen-albert @maho7791 @stbischof - should this remain public, or shall I investigate making it a private implementation detail?

stbischof:
I have to look deeper. But in general it is easy do add a method or class in api in a minor change then removing it an i major.

tcalmant:
I'll propose a PR with the NotificationAccumulator being private to the core.impl bundle

PR #536, opened 2025-01-23
Add a default TimedValue implementation

timothyjward:
The TimedValue interface is an important part of the API, but I noticed an increasing number of _minimal_ implementations that duplicated one another. This isn't a good sign, and means that we should provide a default implementation. As TimedValue is supposed to be immutable and DTO-like this seems like a good time to move to Java 17 and implement the default as a record type. This keeps the API package clean and simple, while allowing us to continue using the TimedValue interface in the API. Where people want or need to create their own TimedValue implementations they can, but the record implementation should serve for most use cases.

### DIFF
--- a/core/api/dependency-check.bndrun
+++ b/core/api/dependency-check.bndrun
@@ -1,6 +1,6 @@
 -runrequires: bnd.identity;id='${project.groupId}.${project.artifactId}'
 -runfw: org.apache.felix.framework
--runee: JavaSE-11
+-runee: JavaSE-17
 
 # These packages are also provided by the gecko.emf.osgi.component project
 # so we blacklist the API for resolve consistency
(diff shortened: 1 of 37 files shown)


timothyjward:
Related to issue #496 and issue #515

core/impl/src/main/java/org/eclipse/sensinact/core/twin/impl/TimedValueImpl.java
tcalmant, line 1: Shouldn't this file be deleted ?

tcalmant (approved):
LGTM
The TimedValueImpl file should be deleted instead of commented out

core/impl/src/main/java/org/eclipse/sensinact/core/twin/impl/TimedValueImpl.java
timothyjward: Good catch

stbischof:
I totally agreed with java 17. We should move!

stbischof (approved):

juergen-albert:
+1 from me

PR #541, opened 2025-02-06
Read retained messages

gathanase:


southbound/mqtt/mqtt-client/src/main/java/org/eclipse/sensinact/gateway/southbound/mqtt/impl/MqttClientHandler.java
```diff
@@ -74,11 +75,14 @@ public class MqttClientHandler implements MqttCallback {
      */
     private Timer reconnectTimer;
 
+    private Object lock = new Object();
```
timothyjward, line 78:
```suggestion
    private final Object lock = new Object();
```

timothyjward (approved):
Overall this looks good. My only comment is in as a suggestion, which is that it's best to make lock objects `final` so that they can be optimised by the JIT, and also so that people don't accidentally change them.

PR #542, opened 2025-02-10
Allow ResourceSelector to compare null values

tcalmant:
Only works for EQUALS and NOT_EQUALS

### DIFF
--- a/filters/resource.selector.impl/src/main/java/org/eclipse/sensinact/filters/resource/selector/impl/ResourceSelectorCriterion.java
+++ b/filters/resource.selector.impl/src/main/java/org/eclipse/sensinact/filters/resource/selector/impl/ResourceSelectorCriterion.java
@@ -134,7 +134,7 @@ public class ResourceSelectorCriterion implements ICriterion {
             OperationType ot = vs.operation == null ? OperationType.EQUALS : vs.operation;
             switch(ot) {
                 case EQUALS:
-                    p = check(getterFunction, vs.value, Objects::equals);
+                    p = check(getterFunction, vs.value, Objects::equals, true);
                     break;
                 case GREATER_THAN:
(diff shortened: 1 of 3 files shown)


timothyjward:
We should definitely tighten up our handling of `null`, but I'm not sure that we should allow checking of it with `EQUALS` or `NOT_EQUALS`. I think that in both those cases `value` is supposed to be set and it's an error not to have it. 

It feels more sensible to have an explicit `IS_NULL` or `IS_NOT_SET`, or possibly just using `IS_SET` with `negate = true`

tcalmant:
I think we have to decide what IS_SET means: currently it means the resource having a non-null value, but I think it should instead be a test if the update timestamp is null or not. Having a data explicitly set to null is a tangible use case.

The current behaviour of IS_SET could then be converted to either a negated IS_NULL or handled with a negated EQUALS "null"

timothyjward:
> I think we have to decide what IS_SET means: currently it means the resource having a non-null value, but I think it should instead be a test if the update timestamp is null or not. Having a data explicitly set to null is a tangible use case.
> 
> The current behaviour of IS_SET could then be converted to either a negated IS_NULL or handled with a negated EQUALS "null"

We have agreed the following:

1. `IS_SET` determines whether a resource is set. When negated this will select things that are not set
2. For all other checks the resource must be set, even when they are negated. An unset resource value will always fail to match.
3. `IS_NOT_NULL` will be added - this will be true when the resource value is set and has a non `null` value. When negated it will be true when the resource is set to the value `null`
4. All other checks will be false if the value is not set, or if it is set to `null`. This is also true when they are negated. This avoids the difficulties around checking for `null` when doing numerical comparisons, or in Equals/Not Equals

timothyjward:
@tcalmant - are you happy with this solution?

PR #563, opened 2025-03-31
Move to record types for gateway notifications

timothyjward:


core/api/src/main/java/org/eclipse/sensinact/core/notification/ResourceActionNotification.java
```diff
@@ -22,9 +22,8 @@
...
  * ACTION/&lt;model&gt;/&lt;provider&gt;/&lt;service&gt;/&lt;resource&gt;
  */
-public class ResourceActionNotification extends AbstractResourceNotification {
-
-    public Instant timestamp;
+public record ResourceActionNotification(String modelPackageUri, String model, String provider,
+        String service, String resource, Instant timestamp) implements AbstractResourceNotification {
 
     @Override
     public String getTopic() {
```
stbischof, line 29: may switch this into constructor that is only one calculated

core/api/src/main/java/org/eclipse/sensinact/core/notification/AbstractResourceNotification.java
```diff
@@ -12,17 +12,17 @@
 **********************************************************************/
 package org.eclipse.sensinact.core.notification;
 
-public abstract class AbstractResourceNotification {
+public interface AbstractResourceNotification {
```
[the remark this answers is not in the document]
stbischof: if we are in right java version i also would use sealed types;-)

core/api/src/main/java/org/eclipse/sensinact/core/notification/LifecycleNotification.java
```diff
@@ -23,13 +23,9 @@
...
  * LIFECYCLE/&lt;model&gt;/&lt;provider&gt;[/&lt;service&gt;[/&lt;resource&gt;]]
  */
-public class LifecycleNotification extends AbstractResourceNotification {
-
-    public Status status;
-
-    public Object initialValue;
-
-    public Map<String, Object> initialMetadata;
+public record LifecycleNotification(String modelPackageUri, String model, String provider,
```
stbischof, line 26: may encapsulate this fields in en extra record, that is then used in all records

String modelPackageUri, String model, String provider,
        String service, String resource,

timothyjward: > may encapsulate this fields in en extra record

Do you mean a default visibility record? I wouldn't want to make it a public type as it's not something that really forms part of the public API, and we don't want people to expect that they'll get an instance. Making it default visibility would then mean all public records implementing methods which delegate to the default visibility type. The interface we have already guarantees that we don't misspell the names of the record components, so I don't really see the value here.

core/api/src/main/java/org/eclipse/sensinact/core/notification/AbstractResourceNotification.java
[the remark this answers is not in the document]
timothyjward: > if we are in right java version i also would use sealed types;-)

Agreed, but I'm not keen to move beyond 17 for the moment, and I don't think anyone else is either

PR #575, opened 2025-05-07
Support pushed updates using Java Records

### LINKED ISSUE #515: Support Java Records

timothyjward:
This commit enhances the AnnotationMapping utility to scan Java Records as well as DTO types. This offers consistent handling whichever mechanism the user chooses to use.

Fixes #515


timothyjward:
I think @juergen-albert will be interested in this one. I worked hard to make sure that all of the pushed EMF type behaviours are the same too.

stbischof (approved):

PR #578, opened 2025-05-19
Add body in http device factory

gathanase:


timothyjward (approved):
Thank you for the contribution.

PR #583, opened 2025-05-21
Sensorthings specification alignments

timothyjward:
This commit fixes two issues with the sensorthings DTOs

* The DataStream and Observation types used Instant to represent fields that are, in fact, time intervals. As Java has no built in type for this the commit adds TimeInterval, including a Jackson parser for the formats permitted by ISO 8601 (start and end, start and duration, duration and end)
* The Datastream observedArea has been loosened to permit any GeoJSON Geometry. This is to fit with the observed behaviour of FROST server (the reference implementation) which does not enforce that the observed area is a Polygon (unlike the specification text)

### DIFF
--- a/northbound/sensorthings/dto/src/main/java/org/eclipse/sensinact/sensorthings/sensing/dto/Datastream.java
+++ b/northbound/sensorthings/dto/src/main/java/org/eclipse/sensinact/sensorthings/sensing/dto/Datastream.java
@@ -14,10 +14,9 @@ package org.eclipse.sensinact.sensorthings.sensing.dto;
 
 import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL;
 
-import java.time.Instant;
 import java.util.Map;
 
-import org.eclipse.sensinact.gateway.geojson.Polygon;
+import org.eclipse.sensinact.gateway.geojson.Geometry;
 
 import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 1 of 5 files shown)


stbischof:
Are you saying that frost server is not spec complient?

timothyjward:
> Are you saying that frost server is not spec complient?

If you read the specification strictly then probably not for this one element (the GeoJSON). For the time interval they are following SensorThings *and* ISO 8601 correctly.

> And why follow the Reference Implementation and not the spec?

We want to use these DTOs to consume SensorThings data as well as to provide it. If we are too strict with our DTOs then we can't read the output of any of the three public SensorThings instances that I've found (which are all using FROST). As stated above the Time Interval changes needed to be fixed for us to follow the specification

stbischof:
I rerun license check that failed with 429

stbischof:
Server is not overloaded.

They calc the rate limit wrong ;)

stbischof:
I totally understand why you are doing this. Just want to be sure that client that strictly follow the rules of the spec can handle this.

timothyjward:
@stbischof - ok to merge?

vwfox:
> I just asked for the polygon part.
> 
> @vwfox can we handle this in Client and ensure that we handle Servers that answer polygon and geometry 

we can handle both FeatureCollection or Plain Polygon

vwfox:
> I just asked for the polygon part.
> 
> @vwfox can we handle this in Client and ensure that we handle Servers that answer polygon and geometry 

we can handle both FeatureCollection or Plain Polygon

stbischof (approved):

PR #585, opened 2025-05-21
Initial version of a Southbound SensorThings connector

timothyjward:
This commit adds support for mirroring SensorThings devices into the sensiNact digital twin. It includes a dynamic model that is used to represent the devices, a polling loop which queries for the latest data, and Records which map the detected updates into the defined model.

There is currently no support for history


timothyjward:
@juergen-albert - please could you have a look through the way that I've built the model for this. It doesn't feel very elegant, and I don't think it's the right way for us to get users to do things either.

core/models/provider/bnd.bnd
```diff
@@ -0,0 +1 @@
+# This is required, so the bnd generator plugin thinks this is a valid bnd  project
```
timothyjward, line 1:
```suggestion
# This is required, so the bnd generator plugin thinks this is a valid bnd  project

```

PR #591, opened 2025-05-27
Support ResourceSelector as a filter in QueryHandler

tcalmant:
Works as a subscription filter
Requires tests


timothyjward:
I’ve added a commit which I think fixes this issue, and without needing API changes or specific handling for the resource selectors. Please take a look and see what you think.

While looking at the subscription DTOs I feel that the notifications aren’t quite right. If we pass a filter then:

1. we should be given an initial notification with the provider snapshots for that filter
2. We should get a notification with a provider snapshot each time one of those provider snapshots changes, 
3. We should get a notification with a provider snapshot when a new provider starts matching
4. We should get an “end match” notification with a provider snapshot when a provider stops matching the filter
5. We should get a “deleted” notification when the provider is gone. 

There are a couple of new API methods that would be needed:
* A “match” on ICriterion for a complete provider snapshot 
* A query on sensinactsession where we pass a provider name and an ICriterion which gives a single snapshot or null

After that the subscription can cache the latest snapshot it has seen and use that to skip updates it has seen (or that aren’t relevant) and to tell when an end match is occurring.

PR #595, opened 2025-06-20
Add a versions rest endpoint

gathanase:
This makes it easier to check all osgi bundle versions.

The endpoint returns the list of all osgi bundles, along with their name, version, git-ref, git-sha.


stbischof:
would expect a endount like `runtime/bundles`

PR #599, opened 2025-06-27
Full resource selectors

timothyjward:
This PR updates the Resource Selectors to make them more powerful, allowing multiple resources to be selected for providers without resorting to ugly REGEX or over-selection. It also better supports queries with the need for `AND` or `OR` filtering.

### DIFF
--- a/docs/source/filters/ResourceSelector.md
+++ b/docs/source/filters/ResourceSelector.md
@@ -7,12 +7,12 @@ A resource selector is a simple type of filter designed to easily select a sensi
 /* Resource Selector */
 {
   "providers": [
-    /* Provider Selection... */
+    /* One or more Provider Selection entries */
     ...
   ],
   "resources": [
-    /* Resource Selection...
-       Note that these must not contain any value selections
+    /* One or more Resource Selection entries
+       Note that these entries must not contain any "value" selections
     */
     ...
   ]
@@ -25,7 +25,7 @@ A resource selector is a simple type of filter designed to easily select a sensi
   "provider": { /* Selection */ },
   "location": { /* Location Selection */ },
   "resources: [
-    /* Resource Selection... */
+    /* One or more Resource Selection entries */
   ]
 }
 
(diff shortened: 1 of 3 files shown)


filters/resource.selector.impl/src/main/java/org/eclipse/sensinact/filters/resource/selector/impl/ProviderSelectionCriterion.java
```diff
@@ -0,0 +1,166 @@
...
+import org.eclipse.sensinact.core.snapshot.ServiceSnapshot;
+import org.eclipse.sensinact.filters.resource.selector.api.ResourceSelector.ProviderSelection;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ProviderSelectionCriterion {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ProviderSelectionCriterion.class);
+
+    private static final Predicate<?> NEVER = x -> Boolean.FALSE;
```
gg-dim, line 34: It would be a good idea to put ‘ALWAYS’ and ‘NEVER’ in the same class.

docs/source/filters/ResourceSelector.md
tcalmant, line 1: We might want to give some examples of the whole Resource Selector. We currently have samples mostly for the value selection.

docs/source/filters/ResourceSelector.md
```diff
@@ -4,30 +4,79 @@ A resource selector is a simple type of filter designed to easily select a sensi
...
 
 ```json
+/* Resource Selector */
 {
+  "providers": [
+    /* Provider Selection... */
+    ...
+  ],
+  "resources": [
+    /* Resource Selection...
```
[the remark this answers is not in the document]
tcalmant: The idea would be to have a clear distinction between selection (with a value filter) and collection (without)

Issue #606, opened 2025-07-10
Configuration PID for sensonrthings seems a bit inconsistent


isalvadori:
The configuration PIDs used in the sensorthings bundles look a bit inconsistent. 
We have, for instance `"sensinact.sensorthings.northbound.rest" ` and `"sensiNact.northbound.sensorthings.mqtt"`
It would be much more user friendly to have a consistent prefix, like `sensinact.sensorthings.northbound` for both of them

PR #608, opened 2025-07-15
Fix inconsistent config pid #606

gg-dim:
reorder pid parts and add documentation


gg-dim:
I revert the PID change but leave/change the Documentation

timothyjward (approved):

PR #612, opened 2025-07-29
Sensinact(EMF)Provider API update

tcalmant:
Added 3 utility methods:

- `SensinactProvider.getResource(String service, String resource)`: direct access to a `SensinactResource` when we already got hold on a `SensinactProvider`. Reduces the number of intermediate object creations
- `SensinactEMFProvider.getProvider(Class<Provider> type)` and `SensinactEMFProvider.getProvider()`: give access to a copy of the underlying Provider. This allows to edit a provider using its type or via EMF reflection before calling `SensinactEMFProvider.update()`

### DIFF
--- a/core/impl/src/main/java/org/eclipse/sensinact/core/twin/impl/SensinactProviderImpl.java
+++ b/core/impl/src/main/java/org/eclipse/sensinact/core/twin/impl/SensinactProviderImpl.java
@@ -162,6 +162,7 @@ public class SensinactProviderImpl extends CommandScopedImpl implements Sensinac
 
     @Override
     public SensinactResource getResource(String service, String resource) {
+        checkValid();
         Objects.requireNonNull(service, "No service name provided");
         Objects.requireNonNull(resource, "No resource name provided");
 


core/impl/src/main/java/org/eclipse/sensinact/core/twin/impl/SensinactProviderImpl.java
```diff
@@ -137,4 +140,40 @@ public SensinactEMFService getOrCreateService(String name, EClass serviceEClass)
...
+            throw new ClassCastException(
+                    "EMF provider %s can't be cast to %s".formatted(getName(), returnType.getName()));
+        }
+
+        return returnType.cast(EcoreUtil.copy(provider));
+    }
+
+    @Override
+    public SensinactResource getResource(String service, String resource) {
+        Objects.requireNonNull(service, "No service name provided");
```
timothyjward, line 165: Missing call to `checkValid`

```suggestion
        checkValid();
        Objects.requireNonNull(service, "No service name provided");
```

timothyjward (changes requested):
One small fix needed (see the comments)

timothyjward (approved):

PR #616, opened 2025-08-07
Query filters on database level

gg-dim:
Limiting the result to 5 elements with the query parameter $top=5, but gathering hundreds or thousands elements from db and filtering at the end on the REST API level, isn't a good idea.
For a better performance I moved the handling of some REST query parameters (skip, top, orderby) to the TimescaleDB history provider.

### DIFF
--- a/northbound/sensorthings/rest.gateway/src/main/java/org/eclipse/sensinact/sensorthings/sensing/rest/impl/HistoryResourceHelper.java
+++ b/northbound/sensorthings/rest.gateway/src/main/java/org/eclipse/sensinact/sensorthings/sensing/rest/impl/HistoryResourceHelper.java
@@ -1,0 +1,125 @@
+/*********************************************************************
+* Copyright (c) 2023 Contributors to the Eclipse Foundation.
+*
+* This program and the accompanying materials are made
+* available under the terms of the Eclipse Public License 2.0
+* which is available at https://www.eclipse.org/legal/epl-2.0/
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*   Kentyou - initial implementation
+**********************************************************************/
+package org.eclipse.sensinact.sensorthings.sensing.rest.impl;
+
+import java.util.ArrayList;
+import java.util.HashMap;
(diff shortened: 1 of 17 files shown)


juergen-albert:
@timothyjward If you find the time, it would be great if you could take a look at this PR.

northbound/sensorthings/rest.gateway/src/main/java/org/eclipse/sensinact/sensorthings/sensing/rest/impl/HistoryResourceHelper.java
```diff
@@ -0,0 +1,125 @@
...
+                    skip = skip + list.value.size();
+                }
+
+            } while (list.value.size() < count && list.value.size() < top);
+        }
+        return list;
+    }
+
+    private static void extractOrderByParameter(UriInfo uriInfo, Map<String, Object> params) {
+        List<String> orderbyParam = uriInfo.getQueryParameters().get("$orderby");
```
timothyjward, line 85: How does this work when they're ordering by different values in the observation, or when they're expanding things and ordering?

northbound/sensorthings/rest.gateway/src/main/java/org/eclipse/sensinact/sensorthings/sensing/rest/impl/HistoryResourceHelper.java
```diff
@@ -0,0 +1,125 @@
...
+
+            int skip = extractSkipParameter(uriInfo, params);
+            int top = extractTopParameter(uriInfo, params, maxResults);
+            extractOrderByParameter(uriInfo, params);
+
+            List<TimedValue<?>> timed;
+            do {
+                params.put("skip", skip);
+
+                timed = (List<TimedValue<?>>) userSession.actOnResource(historyProvider, "history", "rangeFiltered",
```
timothyjward, line 69: This filtering isn't applying the `$filter` (which is really hard!) nor is it ordering relative to the requested column in `$orderby` which means that it is likely to be giving the wrong answer

PR #621, opened 2025-08-20
Fix GeoJson Point null handling

isalvadori:


southbound/history/timescale-provider/src/main/java/org/eclipse/sensinact/gateway/southbound/history/timescale/TimescaleDatabaseWorker.java
```diff
@@ -174,7 +174,8 @@ public void notify(String topic, ResourceDataNotification event) {
                     "(SELECT ST_GeomFromGeoJSON( ? )::geography)");
             String tmpValue;
             if (event.newValue() == null) {
-                tmpValue = "{\"type\":\"Point\", \"coordinates\":[NaN, NaN]}";
+                //Geometry 'Point empty'
+                tmpValue = "{\"type\":\"Point\", \"coordinates\":[]}";
```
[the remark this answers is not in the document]
timothyjward: That makes sense, but I don't think that it should become part of the `geo-json` module if it's really just a way of storing an empty field in the database column

core/geo-json/src/test/java/org/eclipse/sensinact/gateway/geojson/GeoJsonTest.java
```diff
@@ -161,17 +163,25 @@ void testNaN() throws Exception {
                 new TypeReference<List<Geometry>>() {
                 });
 
-        assertEquals(2, geometries.size());
```
timothyjward, line 164: If we're actually saying that we should be strict then isn't `NaN` in these JSON files invalid. Shouldn't we fail to parse anything?

core/geo-json/src/test/java/org/eclipse/sensinact/gateway/geojson/GeoJsonTest.java
```diff
@@ -161,17 +163,25 @@ void testNaN() throws Exception {
...
+        assertEquals(1, geometries.size());
         assertEquals(Point, geometries.get(0).type);
         Point nan = (Point) geometries.get(0);
         assertTrue(Double.isNaN(nan.coordinates.latitude));
         assertTrue(Double.isNaN(nan.coordinates.longitude));
         assertTrue(Double.isNaN(nan.coordinates.elevation));
-        assertEquals(Point, geometries.get(1).type);
```
timothyjward, line 170: This looks like something that should be moved into another test where we assert the behaviour on the invalid (i.e. empty) coordinates array

core/geo-json/src/test/java/org/eclipse/sensinact/gateway/geojson/GeoJsonTest.java
isalvadori: Then we should add an additional check in the CoordinatesDeserializer and throw an exception even when one of the coordinates is NaN, right?

PR #627, opened 2025-09-10
Update sensiNact GeoJSON to use Record types

timothyjward:
This commit updates the GeoJSON types to be Java records. This significantly improves their use by making them immutable, avoiding the need to copy them repeatedly for use in notifications. The inheritance hierarchy has been replaced by interfaces, and the types are sealed to avoid unexpected extensions beyond the GeoJSON specification.

As an extra bonus, we now have proper support for *empty* GeoJSON types (which can be tested with GeoJsonObject#isEmpty()) using empty arrays in serialization, as described in the GeoJSON specification.


timothyjward:
After the discussions in #621 I thought I would prototype the changes needed to move to Java Records, and *properly* handle the specified behaviour for empty GeoJSON objects. This change simplifies usage quite a bit (records are nicer to construct/populate than DTOs). I would be interested in feedback, and whether people would like to see this merged.

timothyjward:
@gg-dim @juergen-albert @tcalmant any thoughts?

tcalmant:
LGTM

gg-dim:
go for it

PR #643, opened 2025-10-15
Introduces ServiceModel annotation to declare Service EClasses

### LINKED ISSUE #625: Mechanism to set Service EClass in UpdateDTOs via String

juergen-albert:
Fixes #625

### DIFF
--- a/docs/source/core/data-model/EMFModel.md
+++ b/docs/source/core/data-model/EMFModel.md
@@ -50,4 +50,335 @@ Linking a provider `B` into provider `A` does not cause any of the linked provid
 
 ## Creating a custom model
 
-TODO: We need input from Jürgen or Mark here
+Creating a custom EMF model allows you to define structured, typed data models for your sensiNact providers. This is particularly useful when you want to enforce strict schemas, reuse model definitions, or work with existing EMF models from other systems.
+
+### Using an existing EMF model
+
+If you already have an EMF model (e.g., from Eclipse Ecore Designer or another EMF-based system), you can use it directly with sensiNact. There are several approaches depending on whether your provider model is static (predefined services) or dynamic (services created at runtime).
+
+#### Static providers with EMF models
+
+For providers with a fixed, predefined structure, you can reference the EMF model classes directly using annotations:
+
+```java
+@Provider("temperatureSensor1")
+public class TemperatureSensorDTO {
+
+    @Model
+    public EClass providerModel = MyModelPackage.Literals.TEMPERATURE_SENSOR;
+
+    @Service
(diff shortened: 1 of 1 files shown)


core/annotation/src/main/java/org/eclipse/sensinact/core/annotation/dto/ServiceModel.java
```diff
@@ -0,0 +1,58 @@
...
+ * <pre>
+ * &#64;Provider(&quot;exampleProvider&quot;)
+ * public class MyDto {
+ *     &#64;Service(&quot;exampleService&quot;)
+ *     &#64;ServiceModel(&quot;MyServiceModelName&quot;)
+ *     &#64;Data
+ *     public String value;
+ * }
+ * </pre>
+ *
```
timothyjward: Can I also use it on a field of type `EClass`?

core/emf-api/src/main/java/org/eclipse/sensinact/core/emf/dto/EMFGenericDto.java
```diff
@@ -29,6 +29,14 @@ public final class EMFGenericDto extends BaseValueDto {
...
     public EClass modelEClass;
 
+    /**
+     * The name of the {@link EClass} of a service. This must be set, when updating
+     * a {@link DynamicProvider}, where a service might not be available yet, it is
+     * an alternative to the serviceEClass and will be overwritten by the
+     * serviceEClass is set..
+     */
+    public String serviceEClassName;
+
```
timothyjward: We should also add this in the `BaseValueDto` with the `@ServiceModel` annotation on it

core/impl/src/main/java/org/eclipse/sensinact/core/extract/impl/AnnotationMapping.java
```diff
@@ -490,19 +505,19 @@ private static Function<Object, EClass> getModelEClassMappingForElement(Class<?>
...
         }
         return mapping;
     }
 
     private static Function<Object, String> getServiceNameMappingForElement(Class<?> clazz, ElementType elementType,
             AnnotatedElement ae) {
-        Function<Object, String> mapping = getAnnotatedNameMapping(clazz, elementType, ae, Service.class,
-                String.class, Set.of(EClass.class, EReference.class));
+        Function<Object, String> mapping = getAnnotatedNameMapping(clazz, elementType, ae, Service.class, String.class,
+                Set.of(EClass.class, EReference.class));
```
timothyjward: Shouldn't these be removed if Service will now only refer to a Service name?

PR #646, opened 2025-10-21
Added Ant call to generated files EOL to LF

tcalmant:
On Windows, EMF generated Java files with CRLF line endings, which are seen as changes by Git.
This PR proposes a solution to force generated files to have LF endings.


timothyjward:
I think the issue is that windows users see a large number of changes whenever they build as the (re)generated code has native platform line endings.

PR #705, opened 2026-03-20
MVP-2083 Migrate to jdk25

gathanase:


timothyjward:
This looks like a good improvement. I think that the test failure is just due to the changed number of bundles in the base launcher now we’re using the non-extension version of spifly. It should be fine to just increase the expected count.

timothyjward:
Thank you for the fix - it's important for sensiNact to run on the latest Java

PR #707, opened 2026-03-31
Upgrade to Jackson 3

tcalmant:
Jackson 3 introduces breaking changes, starting with a package name change.
A part of those changes is described here: [MIGRATING_TO_JACKSON_3.md](https://github.com/FasterXML/jackson/blob/main/jackson3/MIGRATING_TO_JACKSON_3.md)

With this PR:
- Direct uses of Jackson have been upgraded to Jackson 3
- Some modules (ESRI and OIDC) still require Jackson 2. The ESRI module could overwrite the JSON parsing part to avoid this requirement.
- Features have been updated to include Jackson 3 + Jackson 2 when necessary
- Maven properties defining version numbers have been move up in the hierarchy to be shared between implementations and features
- DEPENDENCIES file have been updated


timothyjward:
This looks good to me. I had hoped that JJWT would include the upgrade PR that has been with them for four months, but they seem unsure what to do about building with Java 17, and even then we would need to wait for a release.

The ESRI dependency may be one that we need to manage. There hasn't been any activity for two years.

PR #710, opened 2026-04-20
Fix: snapshot Collection values in notifications to prevent Jackson serialization

pisua:
### Problem
When a resource value is a multi-value link (e.g. an EMF `EDataTypeUniqueEList`), the `ResourceDataNotification` was storing a direct reference to the mutable collection. If the underlying EMF model was mutated between the time `resourceValueUpdate` was called and the time Jackson serialized the notification, this caused a `BasicIndexOutOfBoundsException` during serialization.

### Root Cause
In `AbstractNotificationAccumulatorImpl.createResourceDataNotification`, the `oldValue` and `newValue` were passed directly into the `ResourceDataNotification` record without being snapshotted. Any structural modification of the backing list (e.g. by an EMF list's internal management) before the notification was flushed and serialized could corrupt Jackson's iteration.

### Fix
Added a `snapshotValue(Object)` helper that, when the value is a `Collection`, creates an immutable copy. Non-collection values are passed through unchanged.

### DIFF
--- a/core/geo-json/pom.xml
+++ b/core/geo-json/pom.xml
@@ -15,6 +15,10 @@
       <artifactId>osgi.annotation</artifactId>
     </dependency>
     <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-annotations</artifactId>
+    </dependency>
+    <dependency>
       <groupId>tools.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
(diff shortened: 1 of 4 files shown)


tcalmant (approved):

timothyjward (approved):
Good catch. We should probably set up a code review to see whether we leak known mutable types from the model anywhere else. The resource values are supposed to be safe once they are in the twin.

PR #711, opened 2026-04-20
Handle special case of multi-values when getting resource content

tcalmant:
Session-based access to resource description was calling `SensinactResource.getValue()`, which always returns only one value.

This PR checks if the value is marked as *many*, so that we get the right resource content (singleton or list)


timothyjward (approved):
As the mapping functions are identical I think that the if block could have just contained the creation of the promise? Logically it’s fine though.

timothyjward (approved):

PR #719, opened 2026-05-04
Fail doResourceWork when the resource is missing

tcalmant:
Currently, we return null if a work was allowed but not applied on a missing resource. We now fail with an IllegalArgumentException, as expected by the Query Handler


timothyjward:
The fix looks fine, but shouldn't we add a test? This looks like an easy bug to reintroduce.

timothyjward (approved):

PR #725, opened 2026-05-05
Add session support for filtered subscriptions

timothyjward:
This commit updates the SensinactSession to include a filtered snapshot subscription, allowing callers to supply a callback which will be notified with the initial snapshot state, and then deltas showing changes to the snapshot that would be returned. This should be much more efficient than polling


tcalmant:
Overall LGTM. Dependencies file must be updated.

Issue #745, opened 2026-06-15
Metadata notifications contain the value


timothyjward:
The gateway metadata notifications include the resource value as a metadata property with the key `value`. This is wrong as it isn’t metadata. We should not include it there and if we are going to include it then it should be a top level property.

PR #747, opened 2026-06-17
Avoid leaking mutable collections from the twin

timothyjward:
The digital twin can contain multi-valued resources. These multi-values can leak in several ways:

* Getting the resource
* Value notifications
* Metadata notifications

This commit ensures that the multi-valued content is copied before exposing it. It also removes the value from the metadata map as it has no business being there.


timothyjward:
This forms part of the fix for #745

pisua (approved):

PR #754, opened 2026-07-15
Updated gecko EMF for released Eclipse Fennec EMF

juergen-albert:
Update to Eclipse Fennec EMF release


juergen-albert:
@timothyjward The change is actual minor, but has some implications for anyone using EMF. We finally released our old gecko EMF under its new Home Eclipse Fennec. This means primarily that anyone using the codegenerator must change the generator name from `geckoEMF` to `fennecEMF`. Exisiting model Bundles can still be deployed with the new Fennec EMF. The only thing required would be our Compatiblity bundle.

timothyjward (approved):

PR #755, opened 2026-07-15
Test hardening for history provider rework

gg-dim:
To ensure the [history provider rework](https://github.com/eclipse-sensinact/org.eclipse.sensinact.gateway/blob/feature/history_provider_rework/HISTORY_PROVIDER_REWORK_PLAN.md) I added some tests.

PR #758, opened 2026-08-26
Rework the history provider onto a `HistoryProvider` service with query pushdown

### LINKED ISSUE #634: Sensorthing filters on database level (history provider)
Limiting the result to 5 elements with the query parameter $top=5, but gathering hundreds or thousands elements from db and filtering at the end on the REST API level, isn't a good idea.
For a better performance we should move the handling of REST query parameters (skip, top, orderby, ...) to the TimescaleDB history provider.

The change isn't trivial as the combinations of parameters result in a rather complex query.

I have started the implementation, but get stuck as the complexity explodes. See PR #616

gg-dim:
Closes #634.

## Motivation

The history support exposed queries only as three ACT actions (`single`/`range`/`count`) with a
500-row cap and a magic empty-marker convention, discovered by provider-name string. SensorThings
fetched thousands of rows and applied `$top`/`$skip`/`$orderby` in the REST layer (#634). The
storage schema lost data: only GeoJSON `Point` was stored (everything else fell back to
`toString()`), and numeric values were narrowed to `long`/`double` on read. An earlier pushdown
attempt (#616) died from combinatorial SQL-template explosion; its review asked for pagination,
counting and filtering to be done database-side *consistently* — this rework implements exactly
that.

## What changed

**New contracts** (`southbound/history/history-api`)
- `HistoryProvider` OSGi service (package `...history.provider` @ 1.0.0): the primary consumer
  contract — paged/streamed range queries via a composable `HistoryQuery` (time range, sort order,
  offset/limit, value filter — one `if` per dimension instead of a template matrix), point-in-time
  and first/latest lookups, counts, capability-gated aggregation. Multiple providers coexist,
  selected by the `sensinact.history.provider.name` service property.
- `HistoryStorage` backend SPI + `HistoryIngestFilter` whiteboard (package `...history.storage`
  @ 1.0.0): backends only store and query; event wiring, filtering, housekeeping and the legacy
  facade are shared engine code. `HistoricalQueries` is deprecated but fully retained.

**History engine** (`southbound/history/history-core`, new)
- Tracks `HistoryStorage` services; per backend it wires the typed-event listener, registers the
  `HistoryProvider` and a bit-identical `HistoricalQueries` ACT facade, and manages the synthetic
  twin provider. Configuration changes re-register services — this also structurally fixes the old
  `@Modified` bug that ignored selector changes.
- Configurable historization filters (factory PID `sensinact.history.filter`): include/exclude
  resource selectors plus change-based suppression (`on-change`, `deadband` with absolute or
  percentage threshold and heartbeat interval), applied per backend at runtime.
- Configurable housekeeping (factory PID `sensinact.history.housekeeping`): retention period,
  keep-count per resource, `max.delete` safety cap, scheduling period.

**Unified storage schema** (`southbound/history/timescale-provider`, rewritten)
- One table (`sensinact.history`) with a kind discriminator, exact `NUMERIC` values plus the Java
  type for faithful reads, and JSONB for any GeoJSON object (including `FeatureCollection`) —
  lossless where the old three-table schema was not. PostGIS is no longer required; the TimescaleDB
  extension is optional (hypertable + `time_bucket` when present, `date_bin` on plain PostgreSQL
  14+). The decision and its evaluation evidence are documented in the ADR
  (`docs/source/southbound/history/history-storage-adr.md`).
- Existing databases are migrated automatically on first start; legacy rows keep their historical
  read shape and the old tables are renamed `*_migrated`, not dropped.
- PID and configuration keys are unchanged (`sensinact.history.timescale`).
- A new in-memory backend (`history-inmemory`) serves tests and demo deployments, and a
  storage contract test suite (`history-testsupport`, ~37 tests) runs against every backend.

**SensorThings query pushdown** (`northbound/sensorthings/*`)
- `$top`/`$skip` and time-field `$orderby` become the history query's pagination and ordering; a
  `$filter` that reduces to time-range bounds and (capability permitting) numeric `result`
  comparisons becomes the query's constraints. For such requests `@iot.count` and the
  `@iot.nextLink` chain are computed database-side and cover the full filtered dataset. Everything
  else (e.g. `or`, functions, string comparisons) keeps the in-memory evaluation path — the
  `PAGINATION_APPLIED` request property tells the response filters when a resource method already
  produced the final page.
- The `+1ms` timestamp hack is replaced by an explicit `TimeRange.millisecondOf` lookup, and both
  SensorThings families plus the filter module now bind the `HistoryProvider` service instead of
  looping over ACT calls (the legacy `sensinact.history.provider` name property is still published
  for one release).

## Behavior change to review

Paginated Observation/HistoricalLocation requests now page over the **full dataset from its
start**. Previously pages were cut from a window of the newest `history.results.max` values — on a
4000-row dataset the first page started at row 1000 and the oldest rows were unreachable while
`@iot.count` still reported all of them. The new behavior is the OData-correct one and the point of
#634; the affected IT expectations were updated with a comment. Requests that still fall back to
in-memory evaluation keep the windowed behavior until their filters become pushable.

## Testing

- Every milestone gated on a full `mvn clean verify`; the SensorThings history ITs and the
  behavioral ACT tests in `TimescaleHistoryTest` were kept assertion-compatible throughout, and
  golden-response compatibility for the ACT actions was pinned ahead of the rework (#755).
- New coverage includes the storage contract suite (run against in-memory, plain PostgreSQL 16 and
  TimescaleDB), a legacy-migration IT, docker-free pushdown ITs on the in-memory backend, and unit
  suites for the engine, facade semantics (empty-marker asymmetry, reverse skip, legacy numeric
  narrowing), filters and the `$filter` constraint extractor.
- Two opt-in load tests verify the rework at scale and stay dormant in normal builds:
  - `TimescaleStorageBenchmarkTest` (`-Dhistory.benchmark=true`): 1.1M rows through the production
    storage on both PostgreSQL variants — ingest ~82–90k rows/s, sub-ms point lookups, 1h/30d
    aggregation 471 ms (Timescale) vs 965 ms (plain).
  - `ObservationHistoryLoadTest` (`HISTORY_BENCHMARK=true`): one million observations served over
    the REST API — a correct page at `$skip=500000` (273 ms) is itself the pushdown proof, since
    the in-memory fallback only sees the newest 3000 values.

## Documentation

`docs/source/southbound/history/` is rewritten around the service contract (ACT actions demoted to
a legacy section), including the new filter/housekeeping configuration, the storage ADR and the
migration notes; `SensorthingsRestAccess.md` gains a query pushdown section.

## Out of scope / follow-ups

Housekeeping selector scoping, an in-memory distribution feature, multi-resource queries
(`MULTI_RESOURCE` capability), keyset pagination for very deep paging, and string/boolean value
filter pushdown are deliberately deferred; the contracts stay additive for all of them.


northbound/sensorthings/rest.gateway/src/main/java/org/eclipse/sensinact/sensorthings/sensing/rest/impl/sensinact/HistoryResourceHelperSensinact.java
```diff
@@ -9,141 +9,243 @@
...
+        if (clause.endsWith(" desc")) {
+            order = SortOrder.DESCENDING;
+            clause = clause.substring(0, clause.length() - 5).trim();
+        } else if (clause.endsWith(" asc")) {
+            clause = clause.substring(0, clause.length() - 4).trim();
+        }
+        return timeFields.contains(clause) ? order : null;
+    }
+
+    private static <T extends Self> ResultList<T> pushedDownResultList(HistoryProvider history, HistoryQuery query,
```
pisua: I understand why the query was passed as a parameter here to compute the count, but since it has already been built and executed to retrieve the values, it feels redundant. Would it be better to avoid passing the entire query and instead pass a record containing just the necessary criteria (path, range, valueFilter), or simply call history.getValueCount() right where the query is executed?

pisua (approved):
except the comments added in the PR and the fact that i can't build this branch on my machine it seems to be ok regarding the code modifications. 
i wanted to let someone else check in his side the build (if it's only me or not) and have a another reviews to ensure the impact of this changes.

gg-dim:
> except the comments added in the PR and the fact that i can't build this branch on my machine it seems to be ok regarding the code modifications. i wanted to let someone else check in his side the build (if it's only me or not) and have a another reviews to ensure the impact of this changes.

Does the failure look like package org.gecko.emf.osgi.configurator does not exist in core/models/provider? That's not this branch — core/models/*/src/main/java is git-ignored codegen output, and anything generated before the Fennec EMF migration on master is stale. Deleting those two directories and rebuilding (mvn install -DskipTests=true) fixes it; current master reproduces the same. If you see a different error, please paste it and I'll dig in.

northbound/sensorthings/rest.gateway/src/main/java/org/eclipse/sensinact/sensorthings/sensing/rest/impl/sensinact/HistoryResourceHelperSensinact.java
gg-dim: Thx for the tip. The count is now computed next to getValues(...) (keeping the HistoryPage.totalCount() shortcut for backends that report it), and pushedDownResultList no longer sees the query at all.

pisua:
i've been able to build it on my machine. it's ok for me. i let Tim and Thomas to have a look.

southbound/history/history-api/src/main/java/org/eclipse/sensinact/gateway/southbound/history/provider/HistoryCapability.java
```diff
@@ -0,0 +1,27 @@
...
+ * {@link HistoryProvider#getCapabilities()} before using capability-gated
+ * operations.
+ */
+public enum HistoryCapability {
+    /** {@link HistoryProvider#aggregate(AggregationQuery)} is supported. */
+    AGGREGATION,
+    /** {@link HistoryQuery#valueFilter()} is executed by the backend. */
+    VALUE_FILTERING,
+    /** {@link HistoryPage#totalCount()} may be populated on range queries. */
+    TOTAL_COUNT
```
timothyjward: There is a missing capability here. Not all history providers will support pruning (deletion) e.g. if we call out to an external web service for history then they don't usually permit deletion.

timothyjward:
Overall this looks good to me - it will be a big change, and should be a big improvement. Thank you.

As I've noted, there is a missing storage feature (i.e. whether the storage is read only and cannot support addition/pruning). This is common for external histories such as utility meter readings or weather service data.

I've also seen that a clean checkout is required to build this, when I simply switched branches there was a mess with the gecko -> fennec migration.

southbound/history/history-api/src/main/java/org/eclipse/sensinact/gateway/southbound/history/provider/HistoryCapability.java
gg-dim: Thx, I added the missing capability `HistoryCapability.PRUNING`. The default `prune()` now throws an `UnsupportedHistoryOperationException` and housekeeping skips with a warning.

timothyjward (approved):

PR #760, opened 2026-09-07
Feat :  adr for design improvment http device factory

pisua:
Here the ADR with multiple choices of implementation and explaination of the needs and recommanded one.

### DIFF
--- a/docs/source/southbound/device-factory/adr_strategie_reference_json_parser_parent_nodes.md
+++ b/docs/source/southbound/device-factory/adr_strategie_reference_json_parser_parent_nodes.md
@@ -57,13 +57,13 @@ Input JSON:
 }
 ```
 
-* **Pros:** 
-  * Zero modifications to the core parser architecture for already mapping that don't need parent node value.
+* **Pros:**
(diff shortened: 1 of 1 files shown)


tcalmant (approved):
LGTM, I also have a preference on Option 2.

timothyjward:
I agree that in abstract the preferred order is `2`,`3`,`1`. My only concerns are:

1. Whether it is possible to implement option 2 without breaking existing configurations (i.e. the default scope should still be populated if no scopes are defined). 
2. That we should implement this in a way that it will *also* work for YAML, XML and other structured data. I think we can ignore CSV though.

timothyjward (approved):

pisua:
> stract the preferred order is `2`,`3`,`1`. My only concerns are:
> 
> 1. Whether it is possible to implement option 2 without breaking existing configurations (i.e. the default scope should still be populated if no scopes are defined).

For me, option 2 was imagined without breaking existing configurations. The aim is to allow more complex new configurations while keeping the current ones working, so it stays backward compatible when parsing the JSON file.

PR #761, opened 2026-09-10
Require Maven 3.9.0 and correct documented build prerequisites

gg-dim:
Maven 3.8.7 can't parse .mvn/maven.config (only # comments in it, which 3.8.x treats as CLI args).


timothyjward (approved):

gg-dim:
fixed with #764

PR #765, opened 2026-09-14
Add a pluggable location filtering implementation

timothyjward:
We use location filtering in a number of parts of sensiNact. ESRI geometry works well, but it is not OSGi packaged, nor are they prepared to stay current with their dependencies (e.g. Jackson 3). This commit introduces a service that can be used to make location filters, and includes implementations backed by ESRI (to maintain compatibility if desired) and JTS (an Eclipse library which also does geometry). I am making JTS the default implementation for the northbound REST API as it is approximately twice as fast in our tests. If desired sensorthings can be moved to use the location filtering service (for now it is tightly coupled to ESRI).


distribution/features/timescale-history-provider-feature/src/main/resources/timescale-history-provider-feature.json
```diff
@@ -4,8 +4,7 @@
     { "id": "org.eclipse.sensinact.gateway.southbound.history:history-api:${gateway.version}" },
     { "id": "org.eclipse.sensinact.gateway.southbound.history:history-core:${gateway.version}" },
     { "id": "org.eclipse.sensinact.gateway.southbound.history:timescale-provider:${gateway.version}" },
-    { "id": "org.eclipse.sensinact.gateway.filters:esri.geometry:${gateway.version}" },
-    { "id": "com.fasterxml.jackson.core:jackson-core:${compatibility.jackson.version}" },
+    { "id": "org.eclipse.sensinact.gateway.filters:jts.geometry:${gateway.version}" },
```
tcalmant, line 7: Don't we need jts.core here ?

timothyjward: Good spot - the test passes because we also have `northbound-rest-feature` but it definitely needs to be included

distribution/features/northbound-rest-feature/src/main/resources/northbound-rest-feature.json
```diff
@@ -14,6 +13,8 @@
     { "id": "org.eclipse.sensinact.gateway.northbound.security:security-api:${gateway.version}" },
     { "id": "org.eclipse.sensinact.gateway.northbound.session:session-api:${gateway.version}" },
     { "id": "org.eclipse.sensinact.gateway.northbound.session:session-impl:${gateway.version}" },
+    { "id": "org.locationtech.jts:jts-core:${jts.version}" },
+    { "id": "tools.jackson.module:jackson-module-jakarta-xmlbind-annotations:${jackson.version}" },
```
tcalmant, line 17: `jackson-module-jakarta-xmlbind-annotations` is duplicated

tcalmant (approved):
LGTM, 2 small issues spotted in features.json files.

